PR #2, opened 2020-04-28
add go module

umairidris:
`$ go mod init github.com/GoogleCloudPlatform/healthcare-data-protection-suite`


MartinPetkov:
Should we add all the dependencies in this PR too?

umairidris:
They will be automatically added when we add the code and run go build

PR #5, opened 2020-04-29
Add cloudbuild check for Terraform deletes

MartinPetkov:
This can be used to prevent automation from executing potentially destructive changes.

Uses jq and an image that already has jq installed.


templates/tfengine/components/cicd/variables.tf
```diff
@@ -64,3 +64,9 @@ variable "build_viewers" {
   description = "List of IAM members to grant cloudbuild.builds.viewer role in the devops project to see CICD results"
   default     = []
 }
+
+variable "fail_on_deletes" {
```
umairidris, line 68: Maybe a better name is reject_deletions_in_plan?

MartinPetkov: Done. Changed to "reject_plan_deletions".

templates/tfengine/components/cicd/main.tf
```diff
@@ -174,6 +174,39 @@ resource "google_cloudbuild_trigger" "plan" {
   ]
 }
 
+resource "google_cloudbuild_trigger" "deletecheck" {
```
umairidris, line 177: deletecheck should be delete_check, but you may want to rename this to be consistent with the reject wording

templates/tfengine/components/cicd/main.tf
```diff
@@ -174,6 +174,39 @@ resource "google_cloudbuild_trigger" "plan" {
...
+  disabled = ! var.trigger_enabled
+  provider = google-beta
+  project  = var.devops_project_id
+  name     = "tf-deletecheck"
```
umairidris, line 181: ditto

templates/tfengine/components/cicd/main.tf
MartinPetkov: I think I'd like to stick with "delete check", because that is what it's doing. The "reject" flag is to control its behaviour in the grander scheme of running as a CICD job.

I'll add a separator though.

templates/tfengine/components/cicd/configs/tf-deletecheck.sh
```diff
@@ -0,0 +1,46 @@
...
+reject_plan_deletions="${1}"
+
+# Parse every plan
+found_deletes="false"
+find "$(pwd)" -name 'plan.tfplan' | while read planfile; do
+  plandir="$(dirname ${planfile})"
+  pushd "${plandir}"
+
+  delchanges="$(terraform show -json $(basename ${planfile}) | jq '.resource_changes[].change | select(.actions | index("delete"))')"
+  if [[ ! "${delchanges}" = "" ]]; then
```
umairidris, line 23: should this be ==? or -z?

templates/tfengine/components/cicd/main.tf
MartinPetkov: Done.

templates/tfengine/components/cicd/main.tf
MartinPetkov: Done

templates/tfengine/components/cicd/configs/tf-deletecheck.sh
MartinPetkov: -z is better, yes

templates/tfengine/components/cicd/main.tf
umairidris: how about deletion_check?

MartinPetkov: Sure

PR #3, opened 2020-04-28
Migrate engine

umairidris:
This is a direct port of the engine code from https://github.com/GoogleCloudPlatform/healthcare/tree/master/deploy/engine with a few changes:

- config.NormalizePath call has been removed and config package was not ported.
- There was a subtle bug still with the mergo.Merge call in the template package which prevented empty fields from being passed down the chain of templates. Fixed by adding `mergo.WithOverwriteWithEmptyValue` option.
- Bazel and BUILD files were removed. README updated to use new process.
- Full example now uses BASE helper var as well.
- Status updated to EAP.

The directory structure follows recommendations from https://github.com/golang-standards/project-layout.

Might be easier to see the branch directly: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/tree/engine

PR #6, opened 2020-04-29
Migrate policygen

xingao267:
This is a direct port of the policygen code, runner package from https://github.com/GoogleCloudPlatform/healthcare/tree/master/deploy with a few changes:

- `config.NormalizePath` func moved to internal/pathutil package as `pathutil.Expand`.
- policygen/config.go is combined into policygen/main.go.
- Bazel and BUILD files were removed.
- Status updated to EAP.
- Fake runner code is removed.


MartinPetkov:
Nit: Please update the description to say where this came from.

MartinPetkov:
I see
```
BIN +4.64 MB cmd/policygen/policygen
```
Should this be removed?

cmd/policygen/main.go
```diff
@@ -0,0 +1,286 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"flag"
```
umairidris, line 31: move to block above

cmd/policygen/main.go
```diff
@@ -0,0 +1,286 @@
...
+)
+
+// Config is the struct representing the Policy Generator configuration.
+type Config struct {
+	OrgID           string                 `json:"org_id"`
+	ForsetiPolicies map[string]interface{} `json:"forseti_policies"`
+	GCPOrgPolicies  map[string]interface{} `json:"gcp_organization_policies"`
+}
+
+// TODO: Move this to input config.
```
umairidris, line 55: We should still try to assign to a person or issue if we can

internal/pathutil/pathutil.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/mitchellh/go-homedir"
+)
+
+// Expand expands the path.
+func Expand(path string) (string, error) {
+	path, err := homedir.Expand(path)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Abs(os.ExpandEnv(path))
```
umairidris, line 33: No need to abs

umairidris:
> I see
> 
> ```
> BIN +4.64 MB cmd/policygen/policygen
> ```
> 
> Should this be removed?

Better yet maybe add .gitignore for all the 3 binaries?

internal/pathutil/pathutil.go
MartinPetkov: Why not abs?

umairidris: Relative paths work fine now. We needed the abs stuff when using bazel because relative from the bazel running dir would not be valid

cmd/policygen/main.go
xingao267: Done.

cmd/policygen/main.go
xingao267: Done.

internal/pathutil/pathutil.go
xingao267: Done.

MartinPetkov:
Re-commenting, please add more info in the PR description, like with https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/3

Otherwise LGTM

xingao267:
> Re-commenting, please add more info in the PR description, like with #3
> 
> Otherwise LGTM

missed that. done now.

PR #7, opened 2020-04-29
Add top level CODEOWNERS.

xingao267:


xingao267:
Yea. one is codeowners will auto-populated in the reviewers list I think. And there is one option in the branch protection rule which requires review from code owner which I'm going to enable.

Issue #10, opened 2020-04-29
Setup CICD for toolkit


umairidris:
Start with just running sanity checks and go test across the repo

PR #16, opened 2020-04-30
Add presubmit config.

xingao267:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/10

### DIFF
--- a/build/go_check.sh
+++ b/build/go_check.sh
@@ -23,15 +23,6 @@ if [[ "${f}" ]]; then
   exit 1
 fi
 
-# Check dependencies up-to-date
-go mod download
-d=$(go get -u all 2>&1)
-if [[ "${d}" ]]; then
-  echo "The following dependencies are updated, please run 'go get -u all' locally and commit the change."
-  echo "${d}"
-  exit 1
-fi
-
 go vet ./...
 go build ./...
 go test ./...
--- a/build/go_check.sh
+++ b/build/go_check.sh
@@ -1,0 +1,37 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


MartinPetkov:
The files LGTM but where does it run? Do we also need a `.github/workflows` file, or a cloudbuild.yaml somewhere?

xingao267:
> The files LGTM but where does it run? Do we also need a `.github/workflows` file, or a cloudbuild.yaml somewhere?

Check the Trigger in the Checks. It's running on GCB and it's using the build/presubmit.yaml.

MartinPetkov:
> > The files LGTM but where does it run? Do we also need a `.github/workflows` file, or a cloudbuild.yaml somewhere?
> 
> Check the Trigger in the Checks. It's running on GCB and it's using the build/presubmit.yaml.

I see, thank you.

build/go_check.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Check format
+f=$(gofmt -l .)
+if [[ "${f}" ]]; then
+  echo "The following files are not formatted:"
+  echo "${f}"
+  exit 1
+fi
+
+# Check dependencies up-to-date
```
MartinPetkov, line 26: Maybe remove this block, given that we have Dependabot enabled.

build/presubmit.yaml
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+- name: "gcr.io/cloud-builders/go"
+  entrypoint: "sh"
```
MartinPetkov, line 17: Why not just make the entrypoint be the check script?

xingao267: That doesn't seem to work. It gives me `standard_init_linux.go:211: exec user process caused "exec format error"`

build/go_check.sh
xingao267: Done.

PR #20, opened 2020-04-30
Add Terraform configs format check.

xingao267:
#10

### DIFF
--- a/build/tf_check.sh
+++ b/build/tf_check.sh
@@ -12,21 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-steps:
-- name: "gcr.io/cloud-builders/npm"
-  entrypoint: "sh"
-  args: ["build/md_check.sh"]
-  waitFor: ["-"] # wait for nothing
+#!/bin/bash
+set -e
 
-- name: "gcr.io/cloud-builders/go"
-  entrypoint: "sh"
-  args: ["build/go_check.sh"]
-  waitFor: ["-"] # wait for nothing
+go install ./cmd/tfengine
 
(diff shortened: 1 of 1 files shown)


umairidris:
I am not too sure about this.. the configs themselves should be tested with go_test and fmt should be done as part of the engine binary.

The reason is there are some cases where we can't know what the final indentation etc will look like e.g. when we have optional vars being added.

Would having a go_test for all examples and running fmt on the generated configs automatically fix this?

build/presubmit.yaml
```diff
@@ -16,10 +16,17 @@ steps:
 - name: "gcr.io/cloud-builders/npm"
   entrypoint: "sh"
   args: ["build/md_check.sh"]
+  waitFor: ["-"] # wait for nothing
```
MartinPetkov, line 19: This is new to me, TIL.

xingao267:
> I am not too sure about this.. the configs themselves should be tested with go_test and fmt should be done as part of the engine binary.
> 
> The reason is there are some cases where we can't know what the final indentation etc will look like e.g. when we have optional vars being added.
> 
> Would having a go_test for all examples and running fmt on the generated configs automatically fix this?
Yes. That will fix it. The test in the PR is just temporary so that before we have those implemented, and as we are actively make changes to the templates. I would like to have something very simple just to make sure they are also sanity checked. Because right now, everytime I make changes to the templates, I do that manually locally.

umairidris:
Let me add this as a test after #21. It shouldn't take too long, just need to do a bit of minor refactoring.

xingao267:
sgtm

Issue #13, opened 2020-04-30
Make sure Engine pins every dependency possible.


xingao267:
provider, module

umairidris:
I think we have most of them. As renovate bot sends out more fixes I will update them to pin to patch version.

PR #21, opened 2020-05-01
Templates & general improvements

umairidris:
- use pathutil in main
- merge data and project into one recipe, so now calling it deploys a full "data" project
- allow setting multiple vars and outputs in terraform components
- generalize terragrunt dependency component and call it terragrunt 'leaf'
- add new terraform recipe that can create vars, outputs and terragrunt configs from one place
- add terraform and provider versions #13 
- set terraform remote state in helper terragrunt leaf component


templates/tfengine/recipes/project/data.yaml
```diff
@@ -1,35 +1,36 @@
...
-  data:
-    NAME: "storage_location"
-    TYPE: "string"
-    VALUE: '"{{.STORAGE_LOCATION}}"'
-{{range $i, $_ := .STORAGE_BUCKETS}}
+{{range $i, $_ := get . "STORAGE_BUCKETS"}}
 - name: "bucket-{{.NAME}}"
   component_path: "../../components/project/storage"
-  output_ref: "dir"
+  output_ref: "data_dir"
```
MartinPetkov, line 32: Does this need to be unique? It would be the same output_ref for every bucket in the range.

umairidris: No, this tells the engine where to write all the buckets. In this case they are all in the same data deployment.

examples/tfengine/full.yaml
```diff
@@ -61,6 +61,18 @@ templates:
...
     ORG_POLICIES: {} # Use default org policies.
+    CICD:
+      PROJECT_ID: "example-devops"
+      STATE_BUCKET: "example-state-bucket"
+      REPO_OWNER: "GoogleCloudPlatform"
+      REPO_NAME: "example"
+      BRANCH_REGEX: "master"
+      CONTINUOUS_DEPLOYMENT_ENABLED: true
+      TRIGGER_ENABLED: true
+      REJECT_PLAN_DELETIONS: false
```
xingao267, line 72: this does not apply anymore

templates/tfengine/components/terraform/variables/variables.tf
```diff
@@ -1,6 +1,8 @@
+{{range .VARS}}
```
xingao267, line 1: add - in {{}}?

umairidris: Adding - actually puts no blank lines so I left this in

examples/tfengine/full.yaml
umairidris: Fixed

xingao267:
@googlebot rescan

PR #27, opened 2020-05-02
Use bash as entrypoint to run deletion check script.

xingao267:
The auto-generated tf-deletion-check.yaml file from engine does not
always have execution permission. Use bash as entrypoint is safer.

### DIFF
--- a/templates/tfengine/components/cicd/configs/tf-plan.yaml
+++ b/templates/tfengine/components/cicd/configs/tf-plan.yaml
@@ -33,7 +33,8 @@ steps:
 
   # Check for delete operations as an FYI, it won't fail the build.
   - name: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:a09ad46f513847d31e2cfe20f818c2d64809e7d12695e2e6098900296f6527bf"
-    entrypoint: '/workspace/${_TERRAFORM_ROOT}/cicd/configs/tf-deletion-check.sh'
+    entrypoint: bash
+    args: ["../cicd/configs/tf-deletion-check.sh"]
     dir: "${_TERRAFORM_ROOT}/org"
 
 substitutions:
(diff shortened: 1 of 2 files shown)


templates/tfengine/components/cicd/configs/tf-plan.yaml
```diff
@@ -33,7 +33,8 @@ steps:
 
   # Check for delete operations as an FYI, it won't fail the build.
   - name: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:a09ad46f513847d31e2cfe20f818c2d64809e7d12695e2e6098900296f6527bf"
-    entrypoint: '/workspace/${_TERRAFORM_ROOT}/cicd/configs/tf-deletion-check.sh'
+    entrypoint: bash
+    args: ["../cicd/configs/tf-deletion-check.sh"]
```
MartinPetkov, line 37: I think this can be replaced with `${_TERRAFORM_ORG}` rather than a relative path.

MartinPetkov: Specifically the ".."

MartinPetkov:
Doesn't it still need execution permissions to be executed, even if the entrypoint is bash?

templates/tfengine/components/cicd/configs/tf-plan.yaml
xingao267: Because the dir is ${_TERRAFORM_ORG}/org and I want to avoid hardcoding the /workspace.

xingao267:
> Doesn't it still need execution permissions to be executed, even if the entrypoint is bash?

Actually not. So if you create a simple script to print hello world. And don't chmod +x it. You can't run it directly but you can still use bash to run it.

templates/tfengine/components/cicd/configs/tf-plan.yaml
MartinPetkov: Nvm. This requires a full path (so /workspace), so I think it's actually better to use relative path here (since if "workspace" ever gets renamed, this script won't break since it's inserted automatically when used in "dir:").

xingao267: exactly

PR #29, opened 2020-05-04
Combine Forseti module with Forseti network.

xingao267:
The Forseti network and NAT is only used for Forseti infra. Separating them into two deployments doesn't give much benefits but make the dependency more complicated.

This PR also enables a few more Forseti features.


umairidris:
Is it worth removing the data dep that causes this in the forseti module upstream? I was thinking of removing the forseti network and reusing a network component.

xingao267:
@umairidris yes, that should be fixed as well but requires changes in the upstream code and waiting for a release. I added a TODO to fix that and later reuse the network component. But I think the goal or end result for this and future state will both be not needing that extra disabled field and forseti part can be done in a single deployment.

PR #30, opened 2020-05-04
Add flag for changing the terraform binary path

MartinPetkov:


MartinPetkov:
> One option for handling this is letting the user set the path to the Terraform binary. Then you can just do:
> 
> --terraform_path="terragrunt" to call terragrunt instead. It also lets you point to other terraform/terragrunt paths and versions if needed. WDYT?
> 
> This is also how I have seen folks using Terragrunt with tools like atlantis CICD so it seems to be a common pattern.

That does seem like a much better approach. I'll update to do that.

PR #31, opened 2020-05-04
Importer: Add support for google_container_cluster

MartinPetkov:


internal/terraform/json.go
```diff
@@ -220,20 +214,20 @@ func ReadProviderConfigValues(data []byte, kind, name string) (map[string]interf
 	return result, nil
 }
 
-func resourceProviderConfig(kind string, name string, plan *plan) (ProviderConfig, error) {
-	// Find the provider_config_key for this resource.
+func resourceProviderConfig(kind string, name string, plan *plan) ProviderConfig {
```
umairidris, line 217: optional: It may be a cleaner interface to return (pc *ProviderConfig, ok bool) where ok represents whether the config was found.

MartinPetkov: Yes, that's cleaner, thank you.

internal/tfimport/resources/google_container_cluster.go
```diff
@@ -0,0 +1,46 @@
...
+package resources
+
+import (
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/terraform"
+)
+
+// GkeClusterImporter defines a struct with the necessary information for a GKE cluster to be imported.
+type GkeClusterImporter struct{}
```
umairidris, line 26: GKE

umairidris: Also, optional but if you're only creating importers in this package then it may be more suitable to call this package importer

then do:

var importers = map[string]importer{
	"google_storage_bucket": &importer.StorageBucket{},
	...
}

internal/tfimport/resources/google_container_cluster.go
```diff
@@ -0,0 +1,46 @@
...
+		return "", fmt.Errorf("could not find location in resource change or provider config")
+	}
+
+	name := fromConfigValues("name", rc.Change.After, nil)
+	if name == nil {
+		return "", fmt.Errorf("could not find cluster name in resource change or provider config")
+	}
+
+	return fmt.Sprintf("projects/%v/locations/%v/clusters/%v", project, location, name), nil
+}
```
umairidris, line 46: p.s. isn't there an exists check missing?

MartinPetkov: Not sure what you mean. If the cluster exists, it'll be imported. If it doesn't exist, it will be handled and surfaced as an appropriate log message.

Do you mean checking if the resource has already been imported? If it has, then it wouldn't show up in the plan as a "create" change.

umairidris: "If it doesn't exist, it will be handled and surfaced as an appropriate log message."

OK, was just making sure this case was handled. LGTM then

internal/tfimport/resources/google_container_cluster.go
MartinPetkov: Done, renamed to "GKE" and renamed the package.

internal/tfimport/resources/google_container_cluster.go
MartinPetkov: Yes, it's here: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/cmd/tfimport/main.go#L134-L135

internal/terraform/json_test.go
```diff
@@ -135,27 +135,29 @@ func TestResourceProviderConfig(t *testing.T) {
...
+		// Provider config exists - should not return empty.
 		{"google_storage_bucket", "gcs_tf_bucket", false},
 	}
 	for _, testcase := range tests {
-		config, err := resourceProviderConfig(testcase.kind, testcase.name, p)
-		if testcase.wantErr {
-			if err == nil {
-				t.Errorf("resourceProviderConfig(%q, %q, %v) returned nil err; wanted err != nil", testcase.kind, testcase.name, testPlanPath)
+		config, ok := resourceProviderConfig(testcase.kind, testcase.name, p)
+		isEmpty := cmp.Equal(config, ProviderConfig{}, cmpopts.EquateEmpty())
```
umairidris, line 153: You shouldn't need to check empty anymore. The ok bool should be the only thing you need to check and if ok then do cmp.Diff

MartinPetkov: I thought about this, I think of it as double-checking - if it returns !ok, it should actually also be returning an empty config, so on the chance that someone ignores the ok and just uses the returned config, it's still not going to be partially populated. WDYT?

umairidris: Most go code does not support using the return value if there is a corresponding ok/err field. Given this is a private func it's even less of a concern. I wouldn't bother and simplify code/tests if possible.

MartinPetkov: I think I see what you mean about using ok and diff. I've simplified the test now.

internal/terraform/json_test.go
```diff
@@ -135,28 +135,24 @@ func TestResourceProviderConfig(t *testing.T) {
...
-			if err == nil {
-				t.Errorf("resourceProviderConfig(%q, %q, %v) returned nil err; wanted err != nil", testcase.kind, testcase.name, testPlanPath)
-			}
-		} else {
-			if cmp.Equal(config, ProviderConfig{}, cmpopts.EquateEmpty()) {
-				t.Errorf("resourceProviderConfig(%q, %q, %v) returned empty provider config; want nonempty config", testcase.kind, testcase.name, testPlanPath)
+		config, ok := resourceProviderConfig(testcase.kind, testcase.name, p)
+
+		if !ok {
+			if diff := cmp.Diff(config, ProviderConfig{}, cmpopts.EquateEmpty()); diff != "" {
```
umairidris, line 154: This wasn't exactly what I meant :P but it's fine to leave this in.

The more important case is when it is "ok" (i.e. the provider config exists). I thought you may want to check the provider config against what you expect to get.

MartinPetkov: I've changed it now to check against a specific expected provider config. I was kind of hoping to avoid doing that since it's somewhat complicated to get the types right, but I realized that I could just copy the output of cmp.Diff and use that as the expected output.

PR #38, opened 2020-05-05
Split resources_test into two tests

MartinPetkov:


umairidris:
> Could you expand on what you mean? I want to check both cases, when it's nil and when it's not.

The reason I suggested splitting up the tests into two was to not require a 'wantErr' field and thus not needing to add complexity to the test to check two different cases. In general if tests can be as simple and straightforward at the cost of slight duplication it's better because you have greater confidence.

In the first test we should handle the case where we expect it to be valid. Thus if there is an error you should report a failure. And then you can check any expected values.

In the second test we should handle the case we want an error. Thus, in this case all you need to do is check if there was no error reported and fail.

MartinPetkov:
> > Could you expand on what you mean? I want to check both cases, when it's nil and when it's not.
> 
> The reason I suggested splitting up the tests into two was to not require a 'wantErr' field and thus not needing to add complexity to the test to check two different cases. In general if tests can be as simple and straightforward at the cost of slight duplication it's better because you have greater readability and thus confidence the test itself is working (and also debugging).
> 
> In the first test we should handle the case where we expect it to be valid. Thus if there is an error you should report a failure. And then you can check any expected values.
> 
> In the second test we should handle the case we want an error. Thus, in this case all you need to do is check if there was no error reported and fail.

I think I understand. Please see the latest changes.

PR #42, opened 2020-05-05
Importer: Add google_organization_iam_audit_config

MartinPetkov:


cmd/tfimport/main.go
```diff
@@ -135,8 +135,9 @@ func run() error {
 			log.Printf("Resource %q does not exist, not importing\n", ir.Change.Address)
 
 		// Important to handle this last.
+		// Don't fail right away, import as many resources as possible.
 		default:
-			return fmt.Errorf("import resource %q: %v", ir.Change.Address, err)
+			log.Printf("import resource %q: %v", ir.Change.Address, err)
```
umairidris, line 140: This will cause the binary to exit 0 even if there are errors right?

One option is you can make a list of errors then print that at the end.

MartinPetkov: I actually decided against doing this just now. I think errors should be surfaced and any other expected situations should be handled by the code.

PR #44, opened 2020-05-06
Update importer supported resources

MartinPetkov:


internal/tfimport/importer/google_container_cluster.go
```diff
@@ -22,10 +22,10 @@ import (
 	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/terraform"
 )
 
-// GKECluster defines a struct with the necessary information for a GKE cluster to be imported.
+// GKECluster defines a struct with the necessary information for a google_container_cluster to be imported.
 type GKECluster struct{}
 
-// ImportID returns the GCP project and bucket name for use in importing.
+// ImportID returns the GCP project, location and name for use in importing.
```
umairidris, line 28: Up to you but it would be ok IMO to just have a single shared comment here like

ImportID returns the ID of the resource to use in importing.

Might be easier because we will support a large number of resources. For more special situations like alert policy etc can supplement with more info.

MartinPetkov: Yeah I was wondering about that. I guess my thought was that comments end up as documentation if ever put through GoDoc, so it's nice if it's specific. But it is already getting a bit tedious, so I'll make it the same comment. Links to the resources are already in the README anyway, and those document how to import.

PR #54, opened 2020-05-06
Importer docs: Update to emphasize running in CWD

MartinPetkov:


cmd/tfimport/README.md
```diff
@@ -15,15 +15,23 @@ generated by the Terrarom Engine.
 
 ## Usage
 
+To run against the current working directory:
+
 ```shell
-go install ./cmd/tfimport
-tfimport --input_dir ~/path/to/terraform/configs/dir/
+go install github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd/tfimport
+tfimport
```
umairidris, line 22: To be clearer you can show two different methods

go install ./cmd/tfimport
cd ~/path/to/terraform/configs/dir
tfimport

go run ./cmd/tfimport --input_dir=~/path...

OR

Put a manfile- like section describing the flags. Then you don't need to provide examples for each flag.

cmd/tfimport/main.go
```diff
@@ -36,8 +36,8 @@ import (
 )
 
 var (
-	inputDir      = flag.String("input_dir", ".", "Required. Path to Terraform config.")
-	terraformPath = flag.String("terraform_path", "terraform", "Optional. Name or path to the terraform binary to use.")
+	inputDir      = flag.String("input_dir", ".", "Path to Terraform config.")
```
umairidris, line 39: Path to directory containing Terraform configs

cmd/tfimport/README.md
```diff
@@ -15,32 +15,38 @@ generated by the Terrarom Engine.
...
 ## Usage
 
+### Installing and Running
+
+To install and run (recommended):
+
 ```shell
-go install ./cmd/tfimport
-tfimport --input_dir ~/path/to/terraform/configs/dir/
+go install github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd/tfimport
```
umairidris, line 23: To be clearer I still think we should show a dummy cd to show that we should be in the tf dir or put a comment.

Also, does `github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd/tfimport` fetch the release at HEAD?

MartinPetkov: It does (technically fetches the default branch, "master" in our case).

PR #55, opened 2020-05-07
Update recipe to create a generic 'resources' deployment, fix GKE and update to latest release

umairidris:


examples/tfengine/full.yaml
```diff
@@ -18,6 +18,10 @@ data:
   ORG_ID: "123"
   BILLING_ACCOUNT: "000-000-000"
   STORAGE_LOCATION: "us-central1"
+  CLUSTER_REGION: "us-central1"
```
MartinPetkov, line 21: Can this be overriden per-cluster? It would be useful to not force all clusters to be in the same region.

examples/tfengine/full.yaml
```diff
@@ -18,6 +18,10 @@ data:
   ORG_ID: "123"
   BILLING_ACCOUNT: "000-000-000"
   STORAGE_LOCATION: "us-central1"
```
xingao267, line 20: btw, this storage_location is somewhat confusing and it's not clear what location it actually controls, and it's not setting the location for everything. We should consider removing this and just set the location at each individual block.

umairidris: Would renaming this to STORAGE_BUCKET_LOCATION be clearer? I think it's helpful if you want to avoid having to duplicate the field everywhere.

xingao267: Does it apply to all storage bucket locations? i.e. state bucket, audit log sink bucket, forseti buckets (currently not) and any user created buckets?

xingao267: Or we can go for a way that this is a convenient field to control everything, if users don't override the location in individual data blocks.

umairidris: state and audit log sink bucket yes, i'm not sure about the forseti one, i don't think we pass any location to it yet.

WDYT about supporting both a global field and allowing individual overrides?

xingao267: supporting both a global field and allowing individual overrides sgtm. If we can actually set on location and apply to everything, that will be fantastic, to buckets, sql, datasets, anything. So make it hard for users to set one in us-east1 and one in us-central1 like what we did in mystudies.

One place might need to look more carefully is if the same type of region or zone is supported by all those servces. i.e. the syntax for region of buckets can potentially be different for sql? or some zone is supported by bucket but not in bigquery. I guess we can start with one global attribute for one type of service to be safe.

umairidris: I originally wanted that, but the problem is not every resource is available in every region. For example, healthcare API is not available in us-east1.

Let me try to do global + local override for every resource for now

xingao267: sg

xingao267: we can do that in a separate PR.

examples/tfengine/full.yaml
umairidris: Will do in follow up PR based on Xin's comments below.

PR #63, opened 2020-05-07
support overrides

### LINKED ISSUE #58: Support child block overriding parent fields
Use case:
overriding storage bucket location on individual storage buckets or only for audit logs, etc.

umairidris:
fix #58


MartinPetkov:
I see that several variables were moved from .tfvars to templating from the engine. I personally feel that it's better to keep as many values as possible in .tfvars and the .tf files generic. One use case of the engine is that someone can come and modify the configs after the fact, and it's easier to do this when most configurable/deployment-specific values are in a .tfvars file and not scattered throughout .tf files themselves. Even if not modifying, it's easier to find out which values are deployment-specific by just looking at the .tfvars file.

My preference is that we keep these changes to .tfvars instead of having the engine template them.

umairidris:
> I see that several variables were moved from .tfvars to templating from the engine. I personally feel that it's better to keep as many values as possible in .tfvars and the .tf files generic. One use case of the engine is that someone can come and modify the configs after the fact, and it's easier to do this when most configurable/deployment-specific values are in a .tfvars file and not scattered throughout .tf files themselves. Even if not modifying, it's easier to find out which values are deployment-specific by just looking at the .tfvars file.
> 
> My preference is that we keep these changes to .tfvars instead of having the engine template them.

That was my initial goal as well, but after doing more digging it becomes very difficult to support this from an engine point of view.

For example, it's fine if you want a single location for all resources, but if you want individual resource overrides then it becomes very difficult.

I can revisit this in the future but for now the boundary I had before has shifted a bit and now I only make variables if it doesn't really make sense for it to ever differ. For example all resources are in the same project, or all GKE clusters will permit the same master authorized networks. However, individual GKE clusters can go in different regions, etc so I put them directly in the code.

PR #66, opened 2020-05-07
Move tfengine code to its own package and add basic test

umairidris:


internal/tfengine/tfengine_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+const basePath = "../../examples/tfengine"
+
+// TestExamples is a basic test for the engine to ensure it runs without error on the examples.
+func TestExamples(t *testing.T) {
+	exs := []string{"simple", "full"}
```
MartinPetkov, line 14: Why not list the .yaml files in that directory? Then adding more examples adds auto testing.

internal/tfengine/tfengine_test.go
```diff
@@ -0,0 +1,35 @@
...
+	for _, ex := range exs {
+		t.Run(ex, func(t *testing.T) {
```
MartinPetkov, line 16: Since you're using `t.Run()`, why not try all examples in parallel?

internal/tfengine/tfengine_test.go
umairidris: Done. I initially wanted to reduce the chances of something going wrong and giving a misleading pass but I added a check for at least one example being found so we are testing something.

internal/tfengine/tfengine_test.go
umairidris: I ran a comparison and it doesn't seem to save any time. Given this is a very quick test there isn't much benefit from parallelization. In this case I would prefer the option with fewer lines of code.

PR #67, opened 2020-05-08
Importer: Add google_storage_bucket_iam_member

MartinPetkov:


internal/tfimport/importer/google_storage_bucket_iam_member.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	member, err := fromConfigValues("member", rc.Change.After, nil)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("b/%v %v %v", bucket, role, member), nil
```
umairidris, line 45: Based on the importing doc the b/ is not required, so let's remove it. It looks inconsistent with other resources.

PR #59, opened 2020-05-07
Importer: Add google_container_node_pool

MartinPetkov:
WANT_LGTM=all

PR #68, opened 2020-05-08
Importer: Add google_cloudbuild_trigger

MartinPetkov:


internal/tfimport/importer/google_cloudbuild_trigger.go
```diff
@@ -0,0 +1,41 @@
...
+type CloudBuildTrigger struct{}
+
+// ImportID returns the ID of the resource to use in importing.
+func (b *CloudBuildTrigger) ImportID(rc terraform.ResourceChange, pcv ProviderConfigMap) (string, error) {
+	project, err := fromConfigValues("project", rc.Change.After, pcv)
+	if err != nil {
+		return "", err
+	}
+
+	name, err := fromConfigValues("name", rc.Change.After, pcv)
```
xingao267, line 35: For cloud build triggers, the import id uses the trigger id https://www.terraform.io/docs/providers/google/r/cloudbuild_trigger.html

You probably will need to use gcloud to get the trigger id base on name first.

xingao267: or if the trigger ID is included in the state, you can just use that directly.

MartinPetkov: I noticed that, but I tested and importing using the name works too.

I'm hoping I can get away with using just the name and not adding the code to do network requests, since I would have to use the name to identify the trigger anyway, and if it's not present, I can't import it. Or, I'd have to add interactivity like in #59 , but I want to wait until that's merged and all comments addressed before I implement it elsewhere.

MartinPetkov: > or if the trigger ID is included in the state, you can just use that directly.

The ID is an attribute, not an argument, so if it's there I think the plan won't show it as being "added", so the importer won't consider it.

PR #69, opened 2020-05-08
Importer: Add google_logging_organization_sink

MartinPetkov:


cmd/tfimport/main.go
```diff
@@ -136,13 +137,19 @@ func run() error {
 
 		// Important to handle this last.
 		default:
-			return fmt.Errorf("import resource %q: %v", ir.Change.Address, err)
+			errMsg := fmt.Sprintf("failed to import resource %q: %v\n%v", ir.Change.Address, err, string(output))
+			errors += fmt.Sprintf("%v\n", errMsg)
```
umairidris, line 141: Consider the following:

var errs []string

for _, cc := ... {
   default:
       errs = append(errs, fmt.Sprintf(...))
}

if len(errs) > 0 {
  return fmt.Errorf("failed to import %v resources:\n%v, len(errs), strings.Join(errs, "\n")
}

MartinPetkov: Good idea. I'm not sure how I forgot that strings.Join is a thing.

cmd/tfimport/main.go
```diff
@@ -137,18 +137,16 @@ func run() error {
...
 	}
 
 	if !importedSomething {
 		log.Printf("No resources imported.")
 	}
 
-	if errors != "" {
-		return fmt.Errorf("Found errors during import:\n%v", errors)
+	if len(errs) > 0 {
+		return fmt.Errorf("Failed to import %v resources:\n%v", len(errs), errs)
```
umairidris: errors should be lowercase

PR #70, opened 2020-05-08
Importer: Add google_compute_network

MartinPetkov:


internal/tfimport/importer/google_compute_network.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	name, err := fromConfigValues("name", rc.Change.After, pcv)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("projects/%v/global/networks/%v", project, name), nil
```
umairidris, line 40: Is "global" always consistent for all VPCs?

MartinPetkov: Yes, it seems so. Both the `id` and the import syntax explicitly say "global" and don't mention this changing.

PR #78, opened 2020-05-08
Do not lock state in plan.

xingao267:


MartinPetkov:
What is the implication of not locking? Would the plan be inconsistent sometimes (if it happens in the middle of an apply), or would it just allow plans to happen in parallel?

xingao267:
> What is the implication of not locking? Would the plan be inconsistent sometimes (if it happens in the middle of an apply), or would it just allow plans to happen in parallel?

Yes. it could be inconsistent. So the plan in presubmit is not *always* true but just FYI. But it can be very common that multiple presubmit jobs run at the same time, or at the same time as the post-deployment job. I would like the apply job to actually grab the lock, and plan jobs just take a peek at the state.

PR #87, opened 2020-05-12
Add schema for engine configs, improve documentation

### LINKED ISSUE #74: Support schema for engine config and recipes
https://json-schema.org/

umairidris:
#74

### DIFF
--- a/internal/jsonschema/validate.go
+++ b/internal/jsonschema/validate.go
@@ -1,0 +1,28 @@
+// Package jsonschema defines helpers for JSON schema validation.
+package jsonschema
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/xeipuuv/gojsonschema"
+)
+
+func Validate(schemaJSON []byte, confJSON []byte) error {
+	result, err := gojsonschema.Validate(gojsonschema.NewBytesLoader(schemaJSON), gojsonschema.NewBytesLoader(confJSON))
+	if err != nil {
+		return fmt.Errorf("validate config: %v", err)
+	}
+
+	if len(result.Errors()) == 0 {
+		return nil
+	}
+
+	var sb strings.Builder
+	sb.WriteString("config has validation errors:")
+	for _, err := range result.Errors() {
+		sb.WriteString(fmt.Sprintf("\n- %s: %s", err.Context().String(), err.Description()))
+	}
+	return errors.New(sb.String())
+}
--- a/internal/jsonschema/validate.go
+++ b/internal/jsonschema/validate.go
@@ -1,3 +1,17 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 // Package jsonschema defines helpers for JSON schema validation.
 package jsonschema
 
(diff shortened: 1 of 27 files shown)


internal/tfengine/schema.go
```diff
@@ -0,0 +1,103 @@
...
+        recipe_path:
+          description: Path to a recipe YAML config. Mutually exclusive with 'component_path'.
+          type: string
+
+        output_ref:
+          description: |
+            OutputRef allows this template write its contents to the same dir as another template.
+
+            Example:
+            If output_ref is set to 'foo.bar' then this will look for sibling template 'foo' (defined in this list) and a child template inside 'foo' called 'bar'.
```
MartinPetkov, line 51: I'm still unclear on how `output_ref` functions. How does the child template 'bar' get used in this case?

internal/tfengine/schema.go
```diff
@@ -0,0 +1,103 @@
...
+          type: string
+
+        data:
```
MartinPetkov, line 54: Does this need a type, i.e. type: object? Or is that the default?

internal/tfengine/schema.go
```diff
@@ -0,0 +1,103 @@
...
+
+            Example:
+              - Flag --output_path = /tmp/engine
+              - Field 'output_ref' set to a template which set output_path to './foo'
+              - This field is set to './bar'
+              - The final output path for this template is: '/tmp/engine/foo/bar'.
+          type: string
+
+        flatten:
+          description: |
```
MartinPetkov, line 81: Does flatten recursively flatten? Or only one level? i.e. if the AUDIT key's value was an object, would its keys be flattened too.

internal/tfengine/load.go
```diff
@@ -0,0 +1,77 @@
...
+	if err := json.Unmarshal(cj, c); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func validate(confJSON []byte) error {
+	schemaJSON, err := yaml.YAMLToJSON([]byte(schema))
+	if err != nil {
+		return fmt.Errorf("failed to load schema: %v", err)
```
xingao267, line 58: remove "failed to" as in v3 we are trying to use just something like fmt.Errorf("load schema")

internal/tfengine/load.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	c := new(Config)
+	if err := json.Unmarshal(cj, c); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func validate(confJSON []byte) error {
```
xingao267, line 55: Can some of these logic be moved to template package and can later be reused by policygen?

internal/tfengine/schema.go
umairidris: I added more explanation, let me know if this is clearer. There will also be schemas for each recipe that will list all the output_refs for each recipe.

internal/tfengine/schema.go
umairidris: Only one level, clarified. Good question.

internal/tfengine/load.go
umairidris: Done

internal/tfengine/load.go
umairidris: Done

internal/tfengine/schema.go
```diff
@@ -0,0 +1,111 @@
...
+          type: string
+
+        recipe_path:
+          description: Path to a recipe YAML config. Mutually exclusive with 'component_path'.
+          type: string
+
+        output_ref:
+          description: |
+            Every template outputs to a specific directory. This field allows
+            a template write its contents to the same output dir as another
```
MartinPetkov, line 49: Typo, "**to** write"

umairidris: Done

internal/jsonschema/validate.go
```diff
@@ -0,0 +1,28 @@
+// Package jsonschema defines helpers for JSON schema validation.
```
xingao267, line 1: license

umairidris: Done

PR #91, opened 2020-05-12
make license consistent and add to recipes

umairidris:


MartinPetkov:
We should have some sort of check to make sure we're not missing license files.

umairidris:
Want to file an FR?

PR #99, opened 2020-05-13
fix network component, add network project to full example

umairidris:


templates/tfengine/components/project/compute_networks/main.tf
```diff
@@ -0,0 +1,64 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */ -}}
+
+{{range get . "COMPUTE_NETWORKS"}}
+{{- $has_secondary_ranges := false}}
+module "{{resourceName .NAME}}" {
+  source  = "terraform-google-modules/network/google"
+  version = "~> 2.0"
```
xingao267, line 19: ~> to a patch version

examples/tfengine/full.yaml
```diff
@@ -87,6 +87,20 @@ templates:
   output_ref: "folder-prod.dir"
   data:
     DISPLAY_NAME: "team1"
+- name: "project-example-prod-networks"
+  recipe_path: "{{$BASE}}/folder/project.yaml"
+    output_ref: "folder-prod-team1.dir"
+  data:
+    PROJECT_ID: "example-prod-networks"
+    RESOURCES:
+      COMPUTE_NETWORKS:
```
MartinPetkov, line 96: Why not have a separate "networks" recipe?

templates/tfengine/components/project/compute_networks/main.tf
```diff
@@ -0,0 +1,64 @@
...
+    },
+
+    {{- if has . "SECONDARY_RANGES"}}
+    {{$has_secondary_ranges = true}}
+    {{- end}}
+    {{- end}}
+  ]
+  {{- end}}
+
+  {{- if $has_secondary_ranges}}
```
MartinPetkov, line 47: I don't think you need this check, the default value of `secondary_ranges` is `{}`.

templates/tfengine/components/project/compute_networks/main.tf
```diff
@@ -0,0 +1,64 @@
...
+      subnet_name            = "{{.NAME}}"
+      subnet_ip              = "{{.IP_RANGE}}"
+      {{- if has . "REGION"}}
+      subnet_region          = "{{.REGION}}"
+      {{- else}}
+      subnet_region          = "{{$.COMPUTE_NETWORK_REGION}}"
+      {{- end}}
+      subnet_flow_logs       = true
+      subnets_private_access = true
+
```
xingao267, line 37: extral line

examples/tfengine/full.yaml
umairidris: IMO being able to define a single project and its resources through one recipe is cleaner. Else you would have way too many templates here and it would be very messy.

templates/tfengine/components/project/compute_networks/main.tf
umairidris: Done

templates/tfengine/components/project/compute_networks/main.tf
umairidris: Done

templates/tfengine/components/project/compute_networks/main.tf
umairidris: This was just to produce more concise configs, if there are no secondary ranges then we don't need to set the field to begin with

PR #104, opened 2020-05-14
Importer: Add random_id

MartinPetkov:
Also includes some branched changes from https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/59, namely adding interactivity.

Tested with the Forseti module, specifically this resource: https://github.com/forseti-security/terraform-google-forseti/blob/master/main.tf#L17

Steps:
1. Apply module as normal, with local tfstate.
2. Remove random_id from state.
3. Plan. Notice that many resources are about to be re-created.
4. Run importer. Provide the hex version of the id, taken from one of the GCE VMs' name.
5. Plan. Notice that this time resources are not being recreated.


internal/tfimport/importer/resources.go
```diff
@@ -14,11 +14,32 @@
...
+	"io"
+	"log"
+	"strconv"
+	"strings"
+)
 
 // ProviderConfigMap is a type alias to make variables more readable.
 type ProviderConfigMap map[string]interface{}
 
+// InsufficientInfoErr indicates that we do not have enough information to
```
xingao267, line 29: enough information to ...?

internal/tfimport/importer/resources.go
```diff
@@ -28,3 +49,81 @@ func fromConfigValues(key string, cvs ...ProviderConfigMap) (interface{}, error)
 	}
 	return nil, fmt.Errorf("could not find key %q in resource change or provider config", key)
 }
+
+// userValue asks the user to fill in a value that the importer can't figure out.
```
xingao267, line 53: s/userValue/fromUser

internal/tfimport/importer/resources.go
```diff
@@ -28,3 +49,81 @@ func fromConfigValues(key string, cvs ...ProviderConfigMap) (interface{}, error)
...
+
+		// Choice must be a valid number.
+		choice, err = strconv.Atoi(val)
+		if err != nil {
+			fmt.Printf("Invalid input %q: %v\n", val, err)
+			continue
+		}
+
+		// Choice must be between the first and final choice.
+		if !(choice >= lowest && choice <= highest) {
```
xingao267, line 119: choice < lowest || choice > highest seems more straightforward than having an addition negation here.

internal/tfimport/importer/resources.go
```diff
@@ -28,3 +49,81 @@ func fromConfigValues(key string, cvs ...ProviderConfigMap) (interface{}, error)
...
+	if err != nil {
+		return "", err
+	}
+	val = strings.TrimSpace(val)
+	return val, nil
+}
+
+// userChoice asks the user to select one of the possible choices by index.
+// Returns skip==true if the user decided not to make a choice.
+func userChoice(in io.Reader, choices []string) (string, error) {
```
xingao267, line 87: Do we currently have a use case for this?

internal/tfimport/importer/resources_test.go
```diff
@@ -75,3 +77,82 @@ func TestFromConfigValuesErr(t *testing.T) {
...
+		stdout := os.Stdout
+		os.Stdout = os.NewFile(0, os.DevNull)
+
+		out, err := userValue(strings.NewReader(tc.input))
+
+		// Restore.
+		os.Stdout = stdout
+
+		if err != nil {
+			t.Errorf("userValue(%q) failed: %s", tc.input, err)
```
xingao267, line 107: s/%s/%v

internal/tfimport/importer/resources_test.go
```diff
@@ -75,3 +77,82 @@ func TestFromConfigValuesErr(t *testing.T) {
...
+		} else if out != tc.want {
```
xingao267, line 108: You can just do `if out != tc.want` instead of else if, which is usually what people do in this case.

internal/tfimport/importer/resources_test.go
```diff
@@ -75,3 +77,82 @@ func TestFromConfigValuesErr(t *testing.T) {
...
+		os.Stdout = os.NewFile(0, os.DevNull)
+
+		out, err := userChoice(strings.NewReader(tc.input), choices)
+
+		// Restore.
+		os.Stdout = stdout
+
+		if err != nil {
+			t.Errorf("userChoice(%q) failed: %s", tc.input, err)
+		} else if out != tc.want {
```
xingao267, line 154: same here

internal/tfimport/importer/resources_test.go
```diff
@@ -75,3 +77,82 @@ func TestFromConfigValuesErr(t *testing.T) {
...
+		stdout := os.Stdout
+		os.Stdout = os.NewFile(0, os.DevNull)
+
+		out, err := userChoice(strings.NewReader(tc.input), choices)
+
+		// Restore.
+		os.Stdout = stdout
+
+		if err != nil {
+			t.Errorf("userChoice(%q) failed: %s", tc.input, err)
```
xingao267, line 153: s/%s/%v

cmd/tfimport/main.go
```diff
@@ -70,6 +73,25 @@ func run() error {
...
+		if retry {
+			log.Println("Succeeded in importing some resources, but others failed. Retrying the import, in case dependent values have now been populated.")
+		}
+	}
+
+	return nil
+}
+
+// This function does the full plan and import cycle.
+// If it imported some resources but failed to import others, it will return true for retry. This is a simple way to solve dependencies without having to figure out the graph.
```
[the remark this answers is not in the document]
MartinPetkov: It only repeats when the function returns "true" for "retry", which only happens when we've imported something (made some progress) but failed to import some other things (aren't done yet). If it's just not possible to import everything (in any order), it will stop looping when it fails to import anything.

cmd/tfimport/main.go
```diff
@@ -134,13 +156,21 @@ func run() error {
 		}
 
 		// Handle the different outcomes of the import attempt.
+		var ie *importer.InsufficientInfoErr
 		switch {
 		// err will only be nil when the import succeed.
+		// Import succeeded, print the success output.
 		case err == nil:
 			// Import succeeded.
 			fmt.Println(output)
```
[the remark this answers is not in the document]
MartinPetkov: I actually want to use fmt.Println here. It prints the coloured output from the terraform commands, I find it generally looks better than being wrapped by a log line. Do you think I should use log, for consistency?

internal/tfimport/importer/resources.go
MartinPetkov: Fixed.

internal/tfimport/importer/resources.go
MartinPetkov: Not yet, but I think we will eventually. I'd like to keep it for a little bit until I go through more of the complicated resources, and if we really don't need it, I'll remove it.

internal/tfimport/importer/resources_test.go
MartinPetkov: I wanted to avoid printing both an error and a "output does not equal wanted". I'll change it and use `t.Fatalf()` above.

internal/tfimport/importer/resources.go
xingao267: sg

cmd/tfimport/main.go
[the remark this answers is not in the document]
xingao267: That's ok. Maybe leave a comment for the reasoning.

cmd/tfimport/main.go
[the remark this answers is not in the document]
xingao267: Is `retry` only `true` when some resources is still in progress or some imports actually error out due to a legit reason?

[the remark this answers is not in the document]
MartinPetkov: It's true only when some resources from the plan were imported successfully, and some were importable but failed to import.

cmd/tfimport/main.go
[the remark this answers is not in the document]
MartinPetkov: Done.

cmd/tfimport/main.go
```diff
@@ -70,6 +73,25 @@ func run() error {
 		importRn = &runner.Default{}
 	}
 
+	retry := true
+	for retry {
```
umairidris, line 77: It think this should be capped at max 5 retries

internal/tfimport/importer/resources.go
```diff
@@ -28,3 +49,81 @@ func fromConfigValues(key string, cvs ...ProviderConfigMap) (interface{}, error)
 	}
 	return nil, fmt.Errorf("could not find key %q in resource change or provider config", key)
 }
+
+// userValue asks the user to fill in a value that the importer can't figure out.
+func fromUser(in io.Reader, fieldName string, prompt string, choices []string) (val string, err error) {
+	log.Printf("Could not determine %q automatically\n", fieldName)
+	log.Println(prompt)
+
+	if len(choices) > 0 {
```
umairidris, line 58: Why do we need logic to pick from choices? random_id can't be given a choice. Can we add this only when it's actually needed?

cmd/tfimport/main.go
MartinPetkov: Why do you suggest that? This isn't a "retry until it works" loop, it's more of a "retry as long as we're making some progress".

internal/tfimport/importer/resources.go
MartinPetkov: I was hoping to have a single function to call for "get value from user". I'll try to rearrange the code so you don't always have to pass in choices even if you're not using them.

cmd/tfimport/main.go
umairidris: Ah ok, I see if all resources error on retry then it will stop. That's fine then

internal/tfimport/importer/resources.go
MartinPetkov: Rearranged, PTAL.

internal/tfimport/importer/resources.go
```diff
@@ -28,3 +49,91 @@ func fromConfigValues(key string, cvs ...ProviderConfigMap) (interface{}, error)
...
+
+// fromUser asks the user for the value, without providing choices.
+func fromUser(in io.Reader, fieldName string, prompt string) (val string, err error) {
+	showPrompt(fieldName, prompt)
+	val, err = userValue(in)
+	return parseUserVal(fieldName, val, err)
+}
+
+// fromUser asks the user to choose a value from the available choices.
+func fromUserChoices(in io.Reader, fieldName string, prompt string, choices []string) (val string, err error) {
```
umairidris, line 77: I am still a bit confused by this, what resource is using this? random id only seem to need fromUser + userValue.

MartinPetkov: None yet, but I think some might. I'd like to keep it for now since I already wrote the code, and if I go through several non-trivial resources and none of them require it, I'll remove it.

MartinPetkov: Removed it, but I have it in a separate branch. I can pull it back if needed in the future.

PR #111, opened 2020-05-15
Importer: Add google_service_account_iam_*

MartinPetkov:


internal/tfimport/importer/google_service_account_iam_binding.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/terraform"
+)
+
+// ServiceAccountIAMBinding defines a struct with the necessary information for a google_service_account_iam_binding to be imported.
+type ServiceAccountIAMBinding struct{}
+
+// ImportID returns the ID of the resource to use in importing.
+func (b *ServiceAccountIAMBinding) ImportID(rc terraform.ResourceChange, pcv ProviderConfigMap) (string, error) {
```
umairidris, line 27: Do we individually need to support the other IAM types too like IAM member?

MartinPetkov: Yes, they're imported differently. I'll add those to this PR too, it makes sense.

umairidris: Looks good, but I wonder if there are going to be way too many files in this dir. Maybe merge them to be more top level like google_iam, google_healthcare, google_storage, etc down the line>

MartinPetkov: That's a good idea. I'd like to have all resources from FDA MyStudies working first, then I'll do a pass and combine into one file where appropriate.

PR #112, opened 2020-05-15
Pull policygen logic to its own package and add basic tests.

xingao267:


cmd/policygen/main.go
```diff
@@ -75,213 +58,7 @@ func main() {
 		log.Fatal("--output_dir must be set")
 	}
 
-	if err := run(); err != nil {
+	if err := policygen.Run(*inputConfig, *inputDir, *inputPlan, *inputState, *outputDir); err != nil {
```
umairidris, line 61: Optional: Given these inputs are all strings and fairly similar it may be beneficial to make a struct e.g.

args := &policygen.RunArgs{InputConfig: *inputConfig, ...}
policygen.Run(args)

xingao267: Done

PR #114, opened 2020-05-15
Importer: Add google_secret_manager_secret{_version}

MartinPetkov:


internal/tfimport/importer/google_secret_manager_secret_version.go
```diff
@@ -0,0 +1,34 @@
...
+type SecretManagerSecretVersion struct{}
+
+// ImportID returns the ID of the resource to use in importing.
+func (b *SecretManagerSecretVersion) ImportID(rc terraform.ResourceChange, pcv ProviderConfigMap) (string, error) {
+	secret, err := fromConfigValues("secret", rc.Change.After, pcv)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("%v/versions/latest", secret), nil
```
umairidris, line 33: Does the secret need to be fully qualified path or just the name of the secret?

MartinPetkov: It should be the .id from google_secret_manager_secret. From the documentation, that's supposed to be `projects/{{project}}/secrets/{{secret_id}}`. So I think for secret_version, the "secret" is already the fully-qualified name.

umairidris: got it, thanks

PR #121, opened 2020-05-15
policygen: add schema

xingao267:


internal/policygen/schema.go
```diff
@@ -0,0 +1,70 @@
...
+
+  forseti_policies:
+    description: |
+      Key value pairs passed to Forseti Policy Library constraint templates.
+    type: object
+    additionalProperties: false
+
+  gcp_organization_policies:
+    description: |
+      Key value pairs passed to GCP Organization Policy constraint templates.
```
MartinPetkov, line 46: Since this is optional, maybe explain what will happen if left blank.

xingao267: Done

PR #122, opened 2020-05-15
Importer: Add google_compute_subnetwork_iam_member_*

MartinPetkov:


internal/tfimport/importer/google_compute_subnetwork_iam.go
```diff
@@ -0,0 +1,51 @@
...
+	subnetwork, err := fromConfigValues("subnetwork", rc.Change.After, nil)
+	if err != nil {
+		return "", err
+	}
+
+	// google_compute_subnetwork_iam_policy only takes the subnetwork
+	importID := fmt.Sprintf("%v", subnetwork)
+
+	// google_compute_subnetwork_iam_binding also takes a role
+	role, err := fromConfigValues("role", rc.Change.After, nil)
```
umairidris, line 39: nit: If this error always is just whether the field exists or not then consider returnng an ok bool instead. Otherwise you are ignoring any potential errors. The other way would be to define a helper func that checks if the field doesn't exist and use errors.As/Is for that.

MartinPetkov: Are you suggesting changing the return value of ImportID or fromConfigValues?

umairidris: Based on offline discussion this isn't needed.

Issue #130, opened 2020-05-15
Support project and folder level org policies.


xingao267:

xingao267:
Fixed by https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/131

PR #131, opened 2020-05-15
policygen: use org policy module instead or resource

### LINKED ISSUE #26: use org policy module instead of resource
Replace [google_organization_policy](https://www.terraform.io/docs/providers/google/r/google_organization_policy.html) with [terraform-google-modules/terraform-google-org-policy](https://github.com/terraform-google-modules/terraform-google-org-policy).

xingao267:
Fixes #26 #130


templates/policygen/org_policies/main.tf
```diff
@@ -17,142 +17,207 @@
...
-
-  boolean_policy {
-    enforced = true
-  }
+module "orgpolicy_appengine_disable_code_download" {
+  source  = "terraform-google-modules/org-policy/google"
+  version = "~> 3.0.2"
+
+  policy_for      = var.policy_for
+  organization_id = var.policy_for == "organization" ? var.node_id : null
```
MartinPetkov, line 34: Instead of repeating these everywhere, why not have a `locals` block up top:
```
locals {
  organization_id = var.policy_for == "organization" ? var.node_id : null
  folder_id       = var.policy_for == "folder" ? var.node_id : null
  project_id      = var.policy_for == "project" ? var.node_id : null
}
...
  organization_id = local.organization_id
  folder_id = local.folder_id
  project_id = local.project_id
```

umairidris: I would actually suggest just using templates here.

{{$policy_for := ""}}
{{$res_field := ""}}
{{$res_id := ""}}
{{if eq .POLICY_FOR "organization"}}
  {{$policy_for = "organization"}}
  {{$policy_for := "organization_id"}}
  {{$res_id = .ORG_ID}}
{{else if eq .POLICY_FOR "folder"}}
  ...
{{else}}
  ...
{{end}}

Then in the module:
module "foo" {
  policy_for: "{{$policy_for}}"
  {{$res_field}} = "{{$res_id}}
}

I am going to use this strategy to use a single component to create foundation components for org, folder and project from a single template.

This way is also superior to direct terraform because you don't need to define unneeded fields so you get more concise configs.

umairidris: Also, do the modules offer any benefit apart from letting you set these resources at different levels? If not, maybe we should just stick to the resources here and use our templates to create the different resource types 

e.g.

resource "google_{{.POLICY_FOR}}_policy" "authorized_networks" {
  {{$res_field}} = "{{$res_id}}"
}

umairidris: One more idea you can also pass these values in from the recipes, so the templates become pretty simple.

xingao267: re: module vs. resource: The module also gives the benefits of easily excluding projects or folders, which is a useful feature to be added later. And the interface is actually more readable imo, so I would prefer module here.

xingao267: For the other topic, I think the config as is + Martin's suggestion for the local vars is actually the cleanest way here. I'm not too sure about using heavy template in `main.tf` here, and the `policy_for` and `node_id` now at least apply across the entire `main.tf`. Maybe later if we are treating each policy individually, then we can put heavy templating in the `main.tf`,

Another thing I just thought about is if we are preparing to give the terraform configs to the users, and they might take it from there, they actually follow the terraform's convention to add stuff in variables.tf and terraform.tfvars. Leaving some common or basic variables in our generated `variables.tf` might be helpful.

xingao267: okay, changed my mind. I'm sold on the template way in `main.tf` for this.

umairidris: variables are fine, i just dont see the benefit of having org/folder/project id checks happening in native terraform or putting every field in every resource when we can use templates to generate cleaner ones

xingao267: I tried out the template way, it's pretty clean.

examples/tfengine/simple.yaml
```diff
@@ -57,7 +57,9 @@ templates:
     MONITOR:
       PROJECT_ID: "example-monitor"
       DOMAIN: "example.com"
-    ORG_POLICIES: {} # Use default org policies.
+    ORG_POLICIES:
+      POLICY_FOR: organization
+      NODE_ID: "12345678"
```
umairidris, line 62: Is RESOURCE_ID clearer?

examples/tfengine/simple.yaml
```diff
@@ -57,7 +57,9 @@ templates:
     MONITOR:
       PROJECT_ID: "example-monitor"
       DOMAIN: "example.com"
-    ORG_POLICIES: {} # Use default org policies.
+    ORG_POLICIES:
+      POLICY_FOR: organization
```
umairidris, line 61: Is RESOURCE_TYPE clearer?

templates/policygen/org_policies/main.tf
```diff
@@ -16,143 +16,183 @@
 # (https://cloud.google.com/resource-manager/docs/organization-policy/org-policy-constraints)
 # See the following resources for the details of policies enforced.
 
-terraform {
-  backend "gcs" {}
-}
+{{- $res_field := ""}}
```
umairidris, line 19: How about $res_field := ".RESOURCE_TYPE"_id}}

examples/tfengine/simple.yaml
xingao267: What about ROOT_ID and ROOT_TYPE (or ANCESTRY_ID and ANCESTRY_TYPE). Resource sounds more like a vm, bucket, etc.

templates/policygen/org_policies/main.tf
xingao267: Done.

examples/tfengine/simple.yaml
xingao267: See above comment.

examples/tfengine/simple.yaml
umairidris: ROOT sgtm.

Another option could be PARENT, that's what's used in https://github.com/terraform-google-modules/terraform-google-log-export

xingao267: That's a good example. I'll be consistent with that.

xingao267:
> You can use the terragrunt leaf component which will take care of this for you

what does that mean?

templates/policygen/org_policies/main.tf
```diff
@@ -18,6 +18,10 @@
 
 {{- $type_field := printf "%s_id" .ROOT_TYPE}}
 
+terraform {
```
umairidris, line 21: meant to put the comment here for terragrunt leaf

umairidris: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/tree/master/templates/tfengine/components/terragrunt/leaf

xingao267: Done

examples/tfengine/simple.yaml
```diff
@@ -57,7 +57,9 @@ templates:
     MONITOR:
       PROJECT_ID: "example-monitor"
       DOMAIN: "example.com"
-    ORG_POLICIES: {} # Use default org policies.
+    ORG_POLICIES:
+      PARENT_TYPE: organization
```
MartinPetkov, line 61: Note for later: It would be good if we can support "allowed values" for some fields, such as here. It should only be one of a few possible values, but will still generate the template if it's set to i.e. "martin".

xingao267: Yea, I'm actually going to do that in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/121 once this is merged first.

PR #134, opened 2020-05-15
add support for project dependencies

### LINKED ISSUE #132: Add support for creating dependencies between deployments
For example the apps project needs to depend on the network project

umairidris:
Fix #132


templates/tfengine/recipes/org/project.yaml
```diff
@@ -19,13 +19,18 @@ templates:
...
   output_path: "./project.{{.PROJECT_ID}}"
   {{end}}
-- name: "terragrunt"
-  component_path: "../../components/terragrunt/leaf"
+- name: "terraform"
+  recipe_path: "../terraform/terraform.yaml"
   output_ref: "dir"
   output_path: "./project"
+  flatten: [
+    {{if has . "PROJECT_TERRAFORM"}}
```
xingao267, line 27: Are you planning to put other things under `PROJECT_TERRAFORM`?

umairidris: PROJECT_TERRAFORM will support any field in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/terraform/terraform.yaml

xingao267: Is it specific to project? Maybe PROJECT_TERRAFORM_ADDONS?

templates/tfengine/recipes/folder/project.yaml
```diff
@@ -23,6 +23,11 @@ templates:
         name: '"mock-folder"'
     INPUTS:
       folder_id: "dependency.parent_folder.outputs.name"
+  flatten: [
+    {{if has . "PROJECT_TERRAFORM"}}
+    { key: "PROJECT_TERRAFORM" }
```
MartinPetkov, line 28: How do the single brackets work in this context?

umairidris: I was actually able to get around this because this is currently the only flatten happening for this template.

However, the reason is i had flatten: [ ]. This is the JSON mode and thus requires JSON style objects (e.g { }).

templates/tfengine/recipes/org/project.yaml
umairidris: Updated, how does it look now?

xingao267: lg

PR #137, opened 2020-05-16
remove output ref

### LINKED ISSUE #133: Deprecate output ref
Based on https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/132, for dependencies it is better to be explicit on the exact path we want to depend on rather than making it confusing and trying to overdo helpers. Thus, I don't think output_ref is very valuable anymore.

https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/136 will help reduce duplication in output paths.

umairidris:
Fix #133


templates/tfengine/recipes/org/project.yaml
```diff
@@ -12,32 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{$output_path := get . "OUTPUT_PATH"}}
+{{if not $output_path}}
```
MartinPetkov, line 16: Could this be the default value of `get`?

umairidris: Good idea

PR #138, opened 2020-05-19
Add .gitignore for built binaries

MartinPetkov:


.gitignore
```diff
@@ -0,0 +1,3 @@
+**/policygen
+**/tfengine
```
umairidris, line 2: From what I remember this would cause changes in the tfengine cmd/internal folders to not get picked up. Is that still the case?

MartinPetkov:
Tested like this:
```shell
$ vim cmd/tfengine/README.md  # Make some random change
$ go build cmd/tfengine
$ mv ./tfengine ./cmd/tfengine/
$ git status
On branch gitignore-binaries
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   cmd/tfengine/README.md

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.gitignore
```

.gitignore
MartinPetkov: Let me try

MartinPetkov:
```shell
$ go build ./cmd/tfengine/
$ cp tfengine cmd/tfengine/
$ cp tfengine internal/tfengine/
$ git status
On branch gitignore-binaries
Your branch is up to date with 'origin/gitignore-binaries'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   cmd/tfengine/README.md
	modified:   internal/tfengine/load.go
```
Seems to work

xingao267: Note .gitignore is not retroactive, for existing directories that have been tracked by git, it will not be affected. Can you try creating some new directories with name `tfengine`?

MartinPetkov: It does ignore new directories, and new files created under new directories.

I just pushed a commit to **not** ignore directories with those names.

.gitignore
```diff
@@ -0,0 +1,6 @@
...
+**/tfimport
+!**/policygen/
```
umairidris, line 4: Put a comment above binaries and above these dir exceptions so users can know what each block is doing

MartinPetkov: Done.

MartinPetkov:
> Can you pair each binary with its dir exception? I think that's cleaner to read

Done.

PR #142, opened 2020-05-19
move project fields to PROJECT and remove project/folder prefixes

### LINKED ISSUE #140: Remove project. and folder. prefixes from generated dirs

umairidris:
Fix #140

### DIFF
--- a/examples/tfengine/full.yaml
+++ b/examples/tfengine/full.yaml
@@ -86,15 +86,16 @@ templates:
 
 # Prod folder for team 1.
 - recipe_path: "{{$BASE}}/folder/folder.yaml"
-  output_path: "./live/folder.prod"
+  output_path: "./live/prod"
   data:
     DISPLAY_NAME: "team1"
 
 # Prod central networks project for team 1.
 - recipe_path: "{{$BASE}}/folder/project.yaml"
-  output_path: "./live/folder.prod/folder.team1"
+  output_path: "./live/prod/team1"
   data:
-    PROJECT_ID: "example-prod-networks"
+    PROJECT:
+      PROJECT_ID: "example-prod-networks"
     RESOURCES:
       COMPUTE_NETWORKS:
       - NAME: "example-network"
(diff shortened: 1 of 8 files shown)


examples/tfengine/full.yaml
```diff
@@ -119,15 +121,16 @@ templates:
...
-      HOST_PROJECT_ID: "example-prod-networks"
-      SUBNETS:
-      - NAME: example-subnet
+    PROJECT:
+      PROJECT_ID: "example-prod-apps"
+      APIS:
+      - "container.googleapis.com"
+      SHARED_VPC:
+        HOST_PROJECT_ID: "example-prod-networks"
+        SUBNETS:
```
xingao267, line 132: could you remind me if this creates the subnet or not?

templates/tfengine/recipes/org/project.yaml
```diff
@@ -12,14 +12,16 @@
...
-{{$output_path := get . "OUTPUT_PATH" (printf "./project.%s" .PROJECT_ID)}}
+{{$output_path := get . "OUTPUT_PATH" .PROJECT.PROJECT_ID}}
 
 templates:
 - component_path: "../../components/terragrunt/leaf"
   output_path: "{{$output_path}}/project"
 - name: "project"
   component_path: "../../components/project/project"
   output_path: "{{$output_path}}/project"
+  flatten:
```
xingao267, line 23: why do we flatten the `PROJECT` here but not in other recipes?

umairidris: None of the other templates in this list require the fields within PROJECT (e.g. they get the project ID through a terragrunt reference), so there was no need to pass in PROJECT keys to them

examples/tfengine/full.yaml
umairidris: It only adds permissions on the subnet for this project's servic accounts, the subnet itself is in the network project.

PR #148, opened 2020-05-20
Add SimpleImporter

MartinPetkov:
This is meant to be used for most resources, where they only require values from the plan.

Resources will be migrated after submitting this change. It should make it much easier to add new resources.

### DIFF
--- a/internal/tfimport/importer/resources_test.go
+++ b/internal/tfimport/importer/resources_test.go
@@ -16,6 +16,8 @@ package importer
 
 import (
 	"testing"
+
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/terraform"
 )
 
 var configs = []ProviderConfigMap{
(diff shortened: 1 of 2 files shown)


internal/tfimport/importer/simple_importer.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Get required fields.
+	err := loadFields(i.Fields, fieldsMap, rc.Change.After, pcv)
+	if err != nil {
+		return "", err
+	}
+
+	// Build the template.
+	buf := &bytes.Buffer{}
+	err = template.Must(template.New("").Option("missingkey=error").Parse(i.Tmpl)).Execute(buf, fieldsMap)
```
umairidris, line 57: template.Must is meant for globals, here you can just use remove it

internal/tfimport/importer/resources_test.go
```diff
@@ -75,3 +85,43 @@ func TestFromConfigValuesErr(t *testing.T) {
...
+	}{
+		// Empty.
+		{[]string{}, "", ""},
+
+		// Simple case, insert some template fields.
+		{[]string{"project", "name"}, "projects/{{.project}}/resources/{{.name}}", "projects/myproject/resources/myresourcename"},
+	}
+	for _, tc := range tests {
+		imp := &SimpleImporter{Fields: tc.reqFields, Tmpl: tc.tmpl}
+		got, _ := imp.ImportID(resourceChange, configs[0])
```
umairidris, line 103: check error?

MartinPetkov: There's another set of tests for just if it errors. We don't expect it to fail here, should I still check err?

internal/tfimport/importer/simple_importer.go
MartinPetkov: Done.

internal/tfimport/importer/resources_test.go
MartinPetkov: I looked at the other test and it does check err in the expected-good case, so I added it here.

umairidris: Yeah, the idea is that even in the 'good' case there may be an error if you make an unexpected change so you want to fail the test there. In the 'errors' case if there was no error you would fail.

PR #158, opened 2020-05-21
add check to make sure go deps are up-to-date

xingao267:
Fix #158

### DIFF
--- a/build/go_check.sh
+++ b/build/go_check.sh
@@ -26,6 +26,15 @@ if [[ "${f}" ]]; then
   exit 1
 fi
 
+# Check dependencies up-to-date
+go mod download
+d=$(go get -u all 2>&1)
+if [[ "${d}" ]]; then
+  echo "The following dependencies are updated, please run 'go get -u all' locally and commit the change."
+  echo "${d}"
+  exit 1
+fi
+
 go mod tidy
 go vet ./...
 go build ./...
--- a/build/go_check.sh
+++ b/build/go_check.sh
@@ -30,7 +30,7 @@ fi
 go mod download
 d=$(go get -u all 2>&1)
 if [[ "${d}" ]]; then
-  echo "The following dependencies are updated, please run 'go get -u all' locally and commit the change."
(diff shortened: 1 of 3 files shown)


build/go_check.sh
```diff
@@ -26,6 +26,15 @@ if [[ "${f}" ]]; then
   exit 1
 fi
 
+# Check dependencies up-to-date
+go mod download
```
MartinPetkov, line 30: Is this necessary?

xingao267:
Example failed test without updating deps:

https://console.cloud.google.com/cloud-build/builds/9c0eaef3-859f-42d1-921d-c12afda44697;step=2?project=hcls-data-protection-internal

build/go_check.sh
```diff
@@ -26,6 +26,15 @@ if [[ "${f}" ]]; then
...
+d=$(go get -u all 2>&1)
+if [[ "${d}" ]]; then
```
MartinPetkov, line 32: Do you need `-n` or something like that to check that the string isn't empty? I don't know if empty strings are falsey in bash.

build/go_check.sh
```diff
@@ -26,6 +26,15 @@ if [[ "${f}" ]]; then
...
+  echo "The following dependencies are updated, please run 'go get -u all' locally and commit the change."
```
MartinPetkov, line 33: Nit: I prefer "The following dependencies are **out of date**..."

build/go_check.sh
xingao267: Yea, because in the container, the dependencies are not installed, so without running this first. go get -u all will always print stuff.

build/go_check.sh
xingao267: Yea so empty string is evaluated to false, so no need for that. http://www.linuxintro.org/wiki/Babe#empty_strings

build/go_check.sh
xingao267: Done

umairidris:
Fix #158

PR #159, opened 2020-05-21
upgrade gke cluster version to 9.1

umairidris:


umairidris:
> The changes from https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/releases/tag/v9.1.0 LGTM but is there any particular reason to update this, other than to stay up-to-date?

Yeah no reason, just saw there was an update and wanted to pin to patch range

Issue #86, opened 2020-05-11
Support HCL for configs and recipes


umairidris:
To be consistent with Terraform and Terragrunt we should also try to support hcl for all configs being written. Currently engine configs are typically written in YAML.

The primary complexity here is HCL decoder cannot decode to map[string]interface{} (https://github.com/hashicorp/hcl/issues/291).

Terragrunt works around this by duplicating their structs. The input struct defines the map fields as type cty.Value and then do a manual conversion to map using helpers https://github.com/gruntwork-io/terragrunt/blob/27e0fb53c45258d183f6fbb347b718554f15dbaa/config/cty_helpers.go. Then they are set to the output struct as maps to use later on.

PR #160, opened 2020-05-21
Add support for HCL configs

umairidris:
#86


internal/template/template.go
```diff
@@ -36,8 +36,8 @@ var mergoOpts = []func(*mergo.Config){
 // FlattenInfo describes keys to flatten. If index is not a nil pointer then it is assumed the
 // value is a list that must be flattened at the specific index.
 type FlattenInfo struct {
-	Key   string `json:"key"`
-	Index *int   `json:"index"`
+	Key   string `hcl:"key,attr" json:"key"`
```
MartinPetkov, line 39: Is the hcl tag syntax documented anywhere?

internal/tfengine/config.go
```diff
@@ -0,0 +1,130 @@
...
+	b, err := ctyjson.Marshal(*value, cty.DynamicPseudoType)
+	if err != nil {
+		return nil, err
+	}
+
+	type jsonRepr struct {
+		Value map[string]interface{}
+	}
+
+	var jr jsonRepr
```
MartinPetkov, line 80: Nit: I think this could be just
```
var jr struct {
  ...
}
```

umairidris: while vars are often shortened structs should typically be spelled out for the most part. Since go is a strongly typed language it's easy to see the type of a var so we don't need to be too verbose for no benefit in var names (like in this case we can clearly see jr is a jsonRepr), while structs themselves should try to convey their purpose through name.

A few exceptions I have seen are from the testing package like testing.T. My guess is because it is used so often testing.T is clearly known and understood (no need to do testing.Test which will just add extra chars to everyones test files).

internal/template/template.go
umairidris: https://godoc.org/github.com/hashicorp/hcl2/gohcl

internal/tfengine/config.go
umairidris: I misread this initially, this is a good point! done

internal/template/template.go
umairidris: added a comment

MartinPetkov:
Re-LGTM

internal/tfengine/config.go
```diff
@@ -0,0 +1,131 @@
...
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	buf, err := template.WriteBuffer(string(b), data)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/86): deprecate yaml.
```
xingao267, line 113: Do we really wanna go that far with hcl? I think we can at least support both.

internal/tfengine/config.go
```diff
@@ -0,0 +1,131 @@
...
+	cj := buf.Bytes()
+	if filepath.Ext(path) == ".yaml" {
+		cj, err = yaml.YAMLToJSON(cj)
+		if err != nil {
+			return nil, err
+		}
+		path = "file.json"
```
xingao267, line 120: where is this file.json from?

internal/tfengine/config.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/hashicorp/hcl/v2/hclsimple"
+	"github.com/zclconf/go-cty/cty"
+	ctyjson "github.com/zclconf/go-cty/cty/json"
+)
+
+// Config is the user supplied config for the engine.
+// HCL struct tags are documented at https://pkg.go.dev/github.com/hashicorp/hcl2/gohcl.
+type Config struct {
+	// TODO(https://github.com/hashicorp/hcl/issues/291): Remove the need for DataCty.
+	DataCty *cty.Value             `hcl:"data,optional" json:"-"`
```
xingao267, line 35: Add some docs around DataCty?

internal/tfengine/config.go
umairidris: IMO HCL should be the main supported language as we generate terraform and terragrunt configs, all which use HCL. JSON is compatible with HCL as well so we get that support for free. I don't personally see the benefit of keeping yaml around indefinitely. But I can leave this block in, we just won't formally support it?

internal/tfengine/config.go
umairidris: added comments

internal/tfengine/config.go
umairidris: added comments

internal/tfengine/config.go
umairidris: Removed TODO and replaced with simple comment for now.

PR #161, opened 2020-05-21
remove go get check, rebuild deps list from scratch

umairidris:
./... seems to be more concise, all seems to get a lot of unncessary duplicates.


MartinPetkov:
So I'm really confused by the output here. When I try running the command locally, I see no changes to `go.mod` or `go.sum`, yet it still outputs that two packages are "upgraded", and the versions it points to are already in `go.mod`. They're marked as `// indirect`, which is a clue, but I still don't get it.

I'd suggest changing the check to seeing if something's changed in git, rather than if the `go get` command produced output. You can use this to check if there are unstaged changes:
```shell
if ! git diff-index --quiet HEAD --; then
  echo 'Unstaged changes';
fi
```

PR #164, opened 2020-05-21
Prune unneeded things from go.mod and go.sum

MartinPetkov:
Did this by removing the requires block from go.mod, then running `go get -u ./... && go mod tidy`.


MartinPetkov:
Depends on #161

PR #165, opened 2020-05-21
Check for dependency updates

MartinPetkov:
Can't just parse the output because sometimes it outputs changes without actually changing go.mod or go.sum


build/go_check.sh
```diff
@@ -26,6 +26,25 @@ if [[ "${f}" ]]; then
   exit 1
 fi
 
+# Check dependencies up-to-date
+cp go.mod go.mod.prev
+go mod download
+outdated=$(go get -u ./... 2>&1)
+go_mod_diff=$(diff -u go.mod go.mod.prev)
```
xingao267, line 33: diff will return 1 if there is diff, so might exit here right away.

MartinPetkov: Yes, you're right, it seems to exit 1 even if stored in a variable. I'll ignore the exit code here.

PR #166, opened 2020-05-21
Convert examples to hcl

umairidris:
Depends on #160 

#86


xingao267:
Looks better. Just wondering if these are hcl formatter that can be used to check formatting of these files in presubmit?

umairidris:
> Looks better. Just wondering if these are hcl formatter that can be used to check formatting of these files in presubmit?

We may be able to abuse https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-hclfmt-file for it.

umairidris:
Filed https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/175

Issue #50, opened 2020-05-06
Support folder level cicd


xingao267:
The in the foundation recipe to choose org or folder level

PR #170, opened 2020-05-21
support folder level cicd

xingao267:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/50

### DIFF
--- a/examples/tfengine/full.yaml
+++ b/examples/tfengine/full.yaml
@@ -63,9 +63,11 @@ templates:
       project_id: "example-monitor"
       domain: "example.com"
     org_policies:
-      parent_type: folder
+      parent_type: "organization" # one of `organization`, `folder` or `project`
       parent_id: "12345678"
     cicd:
+      parent_type: "organization" # one of `organization` or `folder`
+      parent_id: "12345678"
       project_id: "example-devops"
       state_bucket: "example-state-bucket"
       repo_owner: "GoogleCloudPlatform"
--- a/examples/tfengine/full.yaml
+++ b/examples/tfengine/full.yaml
@@ -48,6 +48,8 @@ templates:
 # Foundation for the org.
 - recipe_path: "{{$base}}/org/foundation.yaml"
   data:
+    parent_type: "organization" # One of `organization` or `folder`
+    parent_id: "12345678"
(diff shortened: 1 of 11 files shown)


examples/tfengine/simple.yaml
```diff
@@ -58,9 +58,11 @@ templates:
       project_id: "example-monitor"
       domain: "example.com"
     org_policies:
-      parent_type: organization
+      parent_type: "folder" # one of `organization`, `folder` or `project`
```
umairidris, line 61:
```suggestion
      parent_type: "folder" # One of `organization`, `folder` or `project`.
```

examples/tfengine/full.yaml
```diff
@@ -63,9 +63,11 @@ templates:
       project_id: "example-monitor"
       domain: "example.com"
     org_policies:
-      parent_type: folder
+      parent_type: "organization" # one of `organization`, `folder` or `project`
       parent_id: "12345678"
     cicd:
+      parent_type: "organization" # one of `organization` or `folder`
```
umairidris, line 69: I believe you can also just set parent_type and parent_id as a sibling of org_policies/cicd. I think this will be cleaner as typically you will want the same parent for both. We can document in the org schema policies that it can also optionally set parent to project if needed.

examples/tfengine/simple.yaml
umairidris: (and in other places)

xingao267: Done

examples/tfengine/full.yaml
xingao267: Done

internal/tfengine/schema.go
```diff
@@ -53,6 +53,32 @@ properties:
         data:
           descripton: Key value pairs passed to this template.
           type: object
+          properties:
```
umairidris, line 56: We could make this a mega schema, but I was planning on adding a schema for each recipe to keep this more manageable, else you will have every possible field here which will be confusing.

e.g. foundation.yamlschema, etc.

You can add a TODO to move this or just add this when the per-recipe schema is done.

PR #171, opened 2020-05-22
Convert recipes to hcl

umairidris:
#86


templates/tfengine/recipes/folder/folder.hcl
```diff
@@ -12,18 +12,27 @@
...
-      path: "../../folder"
-      mock_outputs:
-        name: '"mock-folder"'
-    inputs:
-      parent: "dependency.parent_folder.outputs.name"
-- component_path: "../../components/folder/folder"
-  output_path: "{{.display_name}}/folder"
-  data:
-    display_name: {{.display_name}}
+templates {
```
MartinPetkov, line 15: Is it possible to rename "templates" to "template" (singular)? I'm not sure what the best practice is here, but to me it reads better when each `template { ... }` defines one template. It was plural before because in yaml it's a list of maps, but in HCL it's a separate declaration each time.

umairidris: Yes I'm going to rename to template "foo" { } in a follow up.

PR #180, opened 2020-05-25
Alphabetize and group resources in importer README

MartinPetkov:


MartinPetkov:
Let me fix the code backticks

MartinPetkov:
Would it be better to just alphabetize all resources?

PR #191, opened 2020-05-25
improve policygen schema

### LINKED ISSUE #190: Support different policy types sharing common attributes.

xingao267:
This supports sharing attributes between different types of policies.

Fixes #190

### DIFF
--- a/internal/policygen/schema.go
+++ b/internal/policygen/schema.go
@@ -20,7 +20,7 @@ title: Policy Generator Config Schema
 additionalProperties: false
 required:
 - template_dir
-- org_id
+- overall
 
 properties:
   template_dir:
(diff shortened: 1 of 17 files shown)


internal/policygen/schema.go
```diff
@@ -29,63 +29,89 @@ properties:
...
-        pattern: ^[0-9]{8,25}$
-
       allowed_policy_member_domains:
         description: |
           See templates/policygen/org_policies/variables.tf.
         type: array
         items:
           type: string
 
+  network:
```
umairidris, line 92: Technically network falls under compute in GCP, but it's probably clearer to separate them. I'll leave it to your judgment

templates/policygen/forseti/org/bigquery_allow_locations.yaml
```diff
@@ -19,7 +19,10 @@ metadata:
 spec:
   severity: high
   match:
-    target: ["organization/{{.org_id}}"]
+    target:
```
umairidris, line 22: You can use the hclField func here

internal/policygen/schema.go
xingao267: Changed to `compute` to stick to https://github.com/GoogleCloudPlatform/magic-modules/tree/master/products

Issue #196, opened 2020-05-26
Add integration tests


umairidris:
We need a way to better test the engine recipes and components.

For now we can use Martin's suggestion to run plan-all on the generated terraform configs. Down the line we should have a real environment that we apply test configs to.

PR #197, opened 2020-05-26
Improve test to plan-all live dir from examples

### LINKED ISSUE #135: GKE clusters fail plan if network hasn't been deployed
GKE module also fetches the network similar to Forseti so if they are in separate deployments the plan will fail on the GKE cluster.

Add a disable for clusters so they are not initially generated.

umairidris:
#196

Also fix #135


internal/tfengine/tfengine_test.go
```diff
@@ -17,10 +17,14 @@ package tfengine
 import (
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"testing"
 )
 
+var backendRE = regexp.MustCompile(`(?s)backend "gcs" {.*?}`)
```
MartinPetkov, line 26: I would leave a comment linking to somewhere explaining the `.*?`. I don't think it's obvious since it's a rarely-encountered regex thing. Also maybe point out that by default it works in multiline mode instead of line-by-line.

internal/tfengine/tfengine_test.go
```diff
@@ -33,7 +37,9 @@ func TestExamples(t *testing.T) {
 	}
 
 	for _, ex := range exs {
+		ex := ex // capture range variable
```
MartinPetkov, line 40: I don't understand what this does, could you elaborate?

internal/tfengine/tfengine_test.go
```diff
@@ -44,10 +50,48 @@ func TestExamples(t *testing.T) {
 				t.Fatalf("tfengine.Run(%q, %q) = %v", ex, tmp, err)
 			}
 
-			// Check for the existence of the bootstrap folder.
-			if _, err := os.Stat(filepath.Join(tmp, "bootstrap")); err != nil {
-				t.Fatalf("os.Stat bootstrap dir = %v", err)
+			// Run plan on live dir to verify configs.
+			path := filepath.Join(tmp, "live")
+
+			convertToLocalBackend(t, path)
```
MartinPetkov, line 56: Maybe add a comment about why this is necessary.

internal/tfengine/tfengine_test.go
```diff
@@ -44,10 +50,48 @@ func TestExamples(t *testing.T) {
...
+			plan.Dir = path
+			if b, err := plan.CombinedOutput(); err != nil {
+				t.Fatalf("command %v in %q: %v\n%v", plan.Args, path, err, string(b))
 			}
 		})
 	}
 }
+
+// convertToLocalBackend converts the Terraform and Terragrunt configs at path to not reference a GCS backend.
+// This allows the test to run plan-all without causing Terraform to lookup the GCS state bucket.
```
MartinPetkov, line 68: I feel this comment belongs above line 56

internal/tfengine/tfengine_test.go
umairidris: Added a link to doc

build/presubmit.yaml
```diff
@@ -23,7 +23,7 @@ steps:
   args: ["build/md_check.sh"]
   id: Markdownlint check
 
-- name: "gcr.io/cloud-builders/go"
+- name: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:a09ad46f513847d31e2cfe20f818c2d64809e7d12695e2e6098900296f6527bf"
```
xingao267, line 26: We can use a newer one here 1dc28cd8c9a01a0a85330e2292035a6661c5adf19917fe1dff0ac43f66397693

umairidris: Done, also updated the templates

PR #198, opened 2020-05-26
Add simple importer with defaults

MartinPetkov:
This makes it easier to implement some resource importers.

Also refactor a bit and allow `map[string]interface{}` into templates, to allow accessing nested fields from configs (ex. the `node_config` field for `google_container_node_pool`).


internal/tfimport/importer/resources.go
```diff
@@ -113,3 +115,20 @@ func loadFields(fields []string, interactive bool, configValues ...ProviderConfi
 
 	return fieldsMap, nil
 }
+
+func fillTemplate(templ string, fieldsMap map[string]string) (string, error) {
```
umairidris, line 119: we have a helper func for this in the template lib if you want to use that instead

internal/tfimport/importer/simple_defaults_importer.go
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package importer
+
+import (
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/terraform"
+)
+
+// SimpleDefaultsImporter defines a struct with the necessary information to import a resource that only depends on fields from the plan.
```
umairidris, line 21: I don't quite understand why this needs to be a separate resource. We can't just add a Defaults field under SimpleImporter?

MartinPetkov: I thought that every other resource would need to provide an empty Defaults. But now I realize that if not provided, it will just be the empty value. I'll adjust the code.

internal/tfimport/importer/resources.go
MartinPetkov: What's the function?

umairidris: The closest is https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/internal/template/template.go#L92, feel free to make adjustments or add new funcs if needed.

internal/tfimport/importer/resources.go
```diff
@@ -16,11 +16,13 @@ package importer
...
+	"bytes"
 	"fmt"
 	"io"
 	"log"
 	"os"
 	"strings"
+	"text/template"
 )
 
 // ProviderConfigMap is a type alias to make variables more readable.
```
xingao267, line 28: Consider adding more documentation around this type. It's getting a bit overloaded as now loadFields can take rc.Change.After, pcv, i.Defaults as ProviderConfigMap.

internal/tfimport/importer/simple_importer.go
```diff
@@ -15,39 +15,24 @@
...
-	// Build the template.
-	buf := &bytes.Buffer{}
-	tmpl, err := template.New("").Option("missingkey=error").Parse(i.Tmpl)
-	if err != nil {
-		return "", err
-	}
-
-	// Execute the template.
-	err = tmpl.Execute(buf, fieldsMap)
+	fieldsMap, err := loadFields(i.Fields, interactive, rc.Change.After, pcv, i.Defaults)
```
xingao267, line 32: Add a bit more documentation for how rc.Change.After, pcv and i.Defaults work together.

internal/tfimport/importer/resources.go
MartinPetkov: Done. Moved the tests over as well.

internal/template/template_test.go
```diff
@@ -122,6 +122,69 @@ func TestMergeData(t *testing.T) {
 		})
 	}
 }
+func TestFillTemplate(t *testing.T) {
```
umairidris, line 125: TestWriteBuffer

internal/template/template_test.go
```diff
@@ -122,6 +122,69 @@ func TestMergeData(t *testing.T) {
...
+					},
+				},
+			},
+			"nested value is: leaf value",
+		},
+	}
+	for _, tc := range tests {
+		buf, err := WriteBuffer(tc.tmpl, tc.fieldsMap)
+		if err != nil {
+			t.Fatalf("fillTemplate(%v, %v) failed: %v", tc.tmpl, tc.fieldsMap, err)
```
umairidris, line 160: WriteBuffer

internal/template/template_test.go
```diff
@@ -122,6 +122,69 @@ func TestMergeData(t *testing.T) {
...
+			t.Errorf("fillTemplate(%v, %v) = %v; want %v", tc.tmpl, tc.fieldsMap, got, tc.want)
+		}
+	}
+}
+
+func TestWriteBufferErr(t *testing.T) {
+	type T map[string]interface{}
+	tests := []struct {
+		tmpl      string
+		fieldsMap T
```
umairidris, line 173: Seems like fieldsMap is never set?

internal/template/template_test.go
```diff
@@ -122,6 +122,69 @@ func TestMergeData(t *testing.T) {
...
+		// Template doesn't use fields.
+		{"mytext", T{"field1": "val1"}, "mytext"},
+
+		// Simple string fields.
+		{"field value is: {{.field1}}", T{"field1": "val1"}, "field value is: val1"},
+
+		// Nested fields.
+		{
+			"nested value is: {{(index .field1 1).inner.leaf}}",
+			T{
```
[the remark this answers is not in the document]
MartinPetkov: Removed it. I just didn't like typing `map[string]interface{}` everywhere.

PR #199, opened 2020-05-26
Make random_id respect the interactive flag

MartinPetkov:


internal/tfimport/importer/random_id.go
```diff
@@ -30,6 +30,9 @@ type RandomID struct{}
 
 // ImportID returns the ID of the resource to use in importing.
 func (c *RandomID) ImportID(rc terraform.ResourceChange, pcv ProviderConfigMap, interactive bool) (string, error) {
+	if !interactive {
+		return "", &InsufficientInfoErr{MissingFields: []string{"b64_url"}}
```
umairidris, line 34: Should we error out or just skip the resource and let a new random ID be created?

MartinPetkov: It will be skipped if this error is returned.

Issue #152, opened 2020-05-20
Add support for Terraform resource based Forseti policies.


xingao267:

xingao267:
Initial set of policies based on Terraform resources are implemented. FR for new policies will be opened individually in the future if needed.

PR #201, opened 2020-05-26
policygen: use tf state file to retrieve resources

xingao267:
- Remove support for using tf plan json file to retrieve resources
- Remove support for using tf state json file to restrieve resources,
this is not as efficient or clean as using native tf state file and
using terraform's own state reading API

#152


internal/policygen/iam.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+type binding struct {
+	Type   string
+	ID     string
+	Role   string
+	Member string
+}
+
+func generateIAMBindingsPolicies(resources []*states.Resource, outputDir string) error {
```
MartinPetkov, line 32: Is the purpose of this function to eventually do more than just call `bindings()`? (I assume yes, from the TODOs, otherwise it should be removed)

internal/policygen/iam.go
```diff
@@ -0,0 +1,86 @@
...
+		"project":      "project",
+		"folder":       "folder",
+		"organization": "org_id",
+	}
+
+	var bindings []*binding
+	for t, idField := range typeToIDField {
+		instances, err := terraform.GetInstancesForType(resources, fmt.Sprintf("google_%s_iam_member", t))
+		if err != nil {
+			return nil, fmt.Errorf("get resource instances for type %q: %v", fmt.Sprintf("google_%s_iam_member", t), err)
```
MartinPetkov, line 56: Nit: This is reused in two places, should just store it in a variable.

internal/policygen/iam.go
```diff
@@ -0,0 +1,86 @@
...
+				ID:     i[idField].(string), // Type checked in validate()
+				Role:   i["role"].(string),
+				Member: i["member"].(string),
+			})
+		}
+	}
+	return bindings, nil
+}
+
+// validate checks the presense of mandatory fields and assert string type.
```
MartinPetkov, line 75: Nit: "presence"

internal/policygen/iam.go
```diff
@@ -0,0 +1,86 @@
...
+func validate(instance map[string]interface{}, mandatoryFields ...string) error {
```
MartinPetkov, line 76: I would suggest making `mandatoryFields` a `[]string`. This way, when calling the function, it's clearer that you're really working with two things (an instance, and a list of mandatory fields that it should contain).

cmd/policygen/main.go
```diff
@@ -27,29 +27,13 @@ import (
 
 var (
 	inputConfig = flag.String("input_config", "", "Path to the Policy Generator config.")
```
umairidris, line 29: With this simplified interface could we be consistent with the engine and do

--config_path
--state_path
--output_path

Also, I thought state is always in JSON? In the future can state path be a GCS path as well where we can recursively parse all states within the bucket? I think that will make it easy to support the engine generated configs.

internal/policygen/iam.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	var bindings []*binding
+	for t, idField := range typeToIDField {
+		instances, err := terraform.GetInstancesForType(resources, fmt.Sprintf("google_%s_iam_member", t))
+		if err != nil {
+			return nil, fmt.Errorf("get resource instances for type %q: %v", fmt.Sprintf("google_%s_iam_member", t), err)
+		}
+
+		for _, i := range instances {
```
umairidris, line 59: Just for clarity, I would rename this to something other than i maybe ins or something. i is usually used for index so can be a bit confusing.

internal/policygen/iam.go
```diff
@@ -0,0 +1,86 @@
...
+
+func generateIAMBindingsPolicies(resources []*states.Resource, outputDir string) error {
+
+	bindings, err := bindings(resources)
+	if err != nil {
+		return err
+	}
+	log.Printf("Found %d bindings from input Terraform resources", len(bindings))
+	// TODO(xingao): Support google_*_iam_bindings.
+	// TODO(xingao): Generate policies.
```
umairidris, line 40: point to issue

internal/policygen/iam.go
xingao267: Yea. This function will also generate the output policy files. Given this PR is already pretty large, I put them as TODOs and will do in following PRs.

internal/policygen/iam.go
xingao267: Done

internal/policygen/iam.go
xingao267: Done

internal/policygen/iam.go
xingao267: Done

internal/policygen/iam.go
xingao267: Done

internal/policygen/iam.go
xingao267: Done

cmd/policygen/main.go
xingao267: Sounds good. Given this PR is already pretty large. I'll do that in a followup PR.

PR #209, opened 2020-05-27
Add all simple resources from the full.hcl example

MartinPetkov:
This list was generated the following way:
1. Run the engine on the full example.
2. Change all backends to "local".
3. Run init to download all modules into .terraform/ folders.
4. Grep for all unique "resource" declarations and filter out ones in `tfimport.go`.

I will add this script to the CI in a follow-up PR.

WANT_LGTM=all

### DIFF
--- a/cmd/tfimport/main.go
+++ b/cmd/tfimport/main.go
@@ -156,9 +156,9 @@ func planAndImport(rn, importRn runner.Runner) (retry bool, err error) {
 				cmd = err.Error()
 			} else {
 				// The last arg in import could be several space-separated strings. These need to be quoted together.
-				args := strings.Split(cmd, " ")
-				if len(args) >= 4 {
-					cmd = fmt.Sprintf("%v %v %v %q\n", args[0], args[1], args[2], strings.Join(args[3:], ""))
+				args := strings.SplitN(cmd, " ", 4)
+				if len(args) == 4 {
+					cmd = fmt.Sprintf("%v %v %v %q\n", args[0], args[1], args[2], args[3])
 				}
 			}
 
--- a/cmd/tfimport/main.go
+++ b/cmd/tfimport/main.go
@@ -151,7 +151,18 @@ func planAndImport(rn, importRn runner.Runner) (retry bool, err error) {
 
 		// In dry-run mode, the output is the command to run.
 		if *dryRun {
-			importCmds = append(importCmds, output)
+			cmd := output
+			if err != nil {
+				cmd = err.Error()
+			} else {
+				// The last arg in import could be several space-separated strings. These need to be quoted together.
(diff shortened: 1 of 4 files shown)


internal/tfimport/tfimport.go
```diff
@@ -186,46 +347,57 @@ var importers = map[string]resourceImporter{
...
+	// Helm provider
+	"helm_release": &importer.SimpleImporter{
+		Fields: []string{"namespace", "name"},
+		Tmpl:   "{{.namespace}}/{{.name}}",
+		Defaults: map[string]interface{}{
+			"namespace": "default",
+		},
+	},
+
+	// Kubernetes provider
```
umairidris, line 386: p.s. looks like the k8s provider is getting replaced with https://github.com/hashicorp/terraform-provider-kubernetes-alpha, so you may just want to support that...

internal/tfimport/importer/random_integer.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return "", err
+	}
+	max, err := fromConfigValues("max", rc.Change.After)
+	if err != nil {
+		return "", err
+	}
+
+	// Seed is optional.
+	seed, _ := fromConfigValues("seed", rc.Change.After)
```
umairidris, line 46: seems a bit smelly to ignore errors, maybe return a specific error type and check for that?

MartinPetkov: I'll replace it with checking if the call returned an error.

internal/tfimport/tfimport.go
MartinPetkov: Where do you see that it's getting replaced?

MartinPetkov: Since it's regarded as experimental, I'll see if I can support both.

cmd/tfimport/main.go
```diff
@@ -151,7 +151,18 @@ func planAndImport(rn, importRn runner.Runner) (retry bool, err error) {
...
 		if *dryRun {
-			importCmds = append(importCmds, output)
+			cmd := output
+			if err != nil {
+				cmd = err.Error()
+			} else {
+				// The last arg in import could be several space-separated strings. These need to be quoted together.
+				args := strings.Split(cmd, " ")
+				if len(args) >= 4 {
+					cmd = fmt.Sprintf("%v %v %v %q\n", args[0], args[1], args[2], strings.Join(args[3:], ""))
```
xingao267, line 161: You are joining the args[3:] with just empty string, is that intentional?

MartinPetkov: Yes, but looking again I actually meant to use `SplitN`. I'll update that.

Issue #212, opened 2020-05-28
Don't enable org policies that will restrict user by default


umairidris:
Couple of org policies by default become deny alls which block common actions.

- member policy domains
- shared vpc host

These should not be generated unless specific fields are set in the template.

xingao267:
Discussed offline. member policy domains policy and allow public vm policy will default to deny all. The rest will be allow all.

PR #215, opened 2020-05-28
Directly using customer ID in domain member org policy

xingao267:


templates/policygen/org_policies/variables.tf
```diff
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-variable "allowed_policy_member_domains" {
-  description = "The list of domains to allow users from, e.g. google.com. If not specified, iam.allowedPolicyMemberDomains constraint will deny all"
+variable "allowed_policy_member_customer_ids" {
+  description = "The list of Cloud Identity or G Suite customer IDs corresponding to the domains to allow users from. If not specified, iam.allowedPolicyMemberDomains constraint will deny all"
```
umairidris, line 16: We should not deny all by default as iam is a very common setting. Can we just not generate this policy?

See #212

xingao267: We can make the default allow all or just don't include that block if the list is empty. But what about other ones? Public VM access, Image project, vpc host project, etc. If we make some allow all and some deny all, it can be confusing. I think it's a broad decision to make what's the default behavior for org_policies {} without any configuration. Forseti policy current has the same behavior. If `allowed_policy_member_domain` is empty then only the gcp service accounts are allowed.

umairidris: We should talk about this a bit more but given the broad scope of org policies I think a lighter touch is perhaps more suitable here. In this case though having this set b default makes sense as a security best practice. Security related ones like public vm access should also always be enabled by default. But some of these like shared vpc projects, image projects can probably be made optional to not over-complicate deployments.

Can you also add a dummy value in the full.hcl and simple.hcl examples?

xingao267: Done for the tf engine examples

PR #216, opened 2020-05-28
add cloud sql instances and users

### LINKED ISSUE #183: Support cloud sql
```
module "my_studies_cloudsql" {
  source  = "GoogleCloudPlatform/sql-db/google//modules/safer_mysql"
  version = "3.2.0"

  name              = "foo"
  project_id        = var.project_id
  region            = var.cloudsql_region
  zone              = var.cloudsql_zone
  availability_type = "REGIONAL"
  database_version  = "MYSQL_5_7"
  vpc_network       = var.network
  user_password     = data.google_secret_manager_secret_version.sql_password.secret_data

  backup_configuration = {
    enabled            = true
    binary_log_enabled = true
    start_time         = "20:55"
  }
}

resource "google_sql_user" "db_user" {
  name     = "db-user"
  instance = module.my_studies_cloudsql.instance_name
  host     = "%"
  password = random_password.db_password.result
  project  = var.project_id
}
```

umairidris:
Fix #183


templates/tfengine/components/project/cloud_sql_instances/main.tf
```diff
@@ -0,0 +1,46 @@
...
+module "{{$instance_resource_name}}" {
+  source  = "GoogleCloudPlatform/sql-db/google//modules/safer_mysql"
+  version = "~> 3.2.0"
+
+  name              = "{{.name}}"
+  project_id        = var.project_id
+  region            = "{{get . "region" $.cloud_sql_region}}"
+  zone              = "{{get . "zone" $.cloud_sql_zone}}"
+  availability_type = "REGIONAL"
+  database_version  = "MYSQL_5_7"
```
MartinPetkov, line 31: I feel we should make this configurable, I image someone would like to use Postgres instead.

umairidris: I added a 'type' that we can use in the future to support postgres.

PR #222, opened 2020-05-29
Importer: Add sanity tests and fix small bugs

### LINKED ISSUE #207: Importer: Add basic sanity tests for all supported importers
These tests would just make sure that, given a sample plan, each importer returns the expected ImportID.

MartinPetkov:
Adding these tests did actually catch a few bugs, so they're worth having, and they're really easy to add.

Fixes #207 

WANT_LGTM=all

### DIFF
--- a/internal/tfimport/tfimport_test.go
+++ b/internal/tfimport/tfimport_test.go
@@ -1015,3 +1015,37 @@ func TestImportersSanity(t *testing.T) {
 		}
 	}
 }
+
+func TestSimpleImporterTmplExtraField(t *testing.T) {
+	imp := &importer.SimpleImporter{
+		Fields: []string{"project", "role"},
+		Tmpl:   "{{.project}} {{.role}} {{.member}}",
+	}
+
+	fields := map[string]interface{}{
+		"project": "my-project",
+		"role":    "roles/owner",
+	}
+	change := terraform.ResourceChange{Change: terraform.Change{After: fields}}
+	_, err := imp.ImportID(change, nil, false)
+	if err == nil {
+		t.Errorf("importer %v ImportID(%v, nil, false) succeeded for malformed input, want error", imp, change)
+	}
+}
+
+func TestSimpleImporterRequiredFieldMissing(t *testing.T) {
+	imp := &importer.SimpleImporter{
+		Fields: []string{"project", "role", "member"},
+		Tmpl:   "{{.project}} {{.role}} {{.member}}",
+	}
+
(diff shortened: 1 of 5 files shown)


internal/tfimport/tfimport_test.go
```diff
@@ -201,3 +202,816 @@ func TestDoesNotExist(t *testing.T) {
...
+			"google_app_engine_application",
+			map[string]interface{}{
+				"project": "my-appengine-project",
+			},
+			"my-appengine-project",
+		},
+		{
+			"google_bigquery_dataset",
+			map[string]interface{}{
+				"project":    "my-bq-project",
```
xingao267, line 223: Do you currently have a test for the ImportID that if one of the fields it's looking for is missing, then it will error out?

MartinPetkov: I don't, I'll add one, good idea.

internal/tfimport/tfimport_test.go
```diff
@@ -201,3 +202,816 @@ func TestDoesNotExist(t *testing.T) {
...
+	for _, tc := range tests {
+		importer, ok := importers[tc.resource]
+		if !ok {
+			t.Fatalf("importer does not exist for %v", tc.resource)
+		}
+
+		change := terraform.ResourceChange{Change: terraform.Change{After: tc.planFields}}
+		got, err := importer.ImportID(change, nil, false)
+		if err != nil {
+			t.Errorf("%v importer %T(%v, nil, false) returned error: %v", tc.resource, importer, change, err)
```
xingao267, line 1011: You can use Fatalf here as the function just errors out

MartinPetkov: Done.

internal/tfimport/tfimport_test.go
```diff
@@ -1015,3 +1015,37 @@ func TestImportersSanity(t *testing.T) {
 		}
 	}
 }
+
+func TestSimpleImporterTmplExtraField(t *testing.T) {
+	imp := &importer.SimpleImporter{
+		Fields: []string{"project", "role"},
+		Tmpl:   "{{.project}} {{.role}} {{.member}}",
```
xingao267, line 1022: Add a note that the template is looking for `member`. And also below that it's missing `member`. It's not very obvious.

MartinPetkov: I'll add a comment above the test that this is what it's testing.

PR #224, opened 2020-05-29
Update CI to check go.sum

MartinPetkov:


build/go_check.sh
```diff
@@ -26,26 +26,45 @@ if [[ "${f}" ]]; then
...
 ${go_mod_diff}
 
-Please run 'go get -u ./...' locally and commit the changes.
 EOF
-  exit 1
 fi
 rm go.mod.prev
 
-go mod tidy
+# Check go.sum as well
```
xingao267, line 50: why do we need to check go.sum? If we do a go get -u ./... and then go mod tidy, should that clean up go.sum and make it deterministic across envs?

umairidris: +1. Let's just stick to go get + go mod tidy as that is the recommendation per go docs

MartinPetkov: I initially thought that go mod tidy wasn't removing values from go.sum. It seemed like it would just add new checksums, but not remove old ones. I did some more experimenting and it seems like it does, so I'll close this PR.

PR #225, opened 2020-05-29
support csr trigger in cicd for testing purposes

### LINKED ISSUE #223: support csr triggers
This feature should be used for testing only.

xingao267:
Fixes #223

### DIFF
--- a/templates/tfengine/components/cicd/manual/main.tf
+++ b/templates/tfengine/components/cicd/manual/main.tf
@@ -156,13 +156,20 @@ resource "google_cloudbuild_trigger" "validate" {
     "${local.terraform_root_prefix}**",
   ]
 
+  {{if index . "github" -}}
   github {
-    owner = var.repo_owner
-    name  = var.repo_name
+    owner = "{{.github.owner}}"
+    name  = "{{.github.name}}"
     pull_request {
-      branch = var.branch_regex
+      branch = "{{.branch_regex}}"
     }
   }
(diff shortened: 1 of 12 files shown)


templates/tfengine/components/cicd/manual/main.tf
```diff
@@ -156,13 +156,20 @@ resource "google_cloudbuild_trigger" "validate" {
...
-    owner = var.repo_owner
-    name  = var.repo_name
+    owner = "{{.github.owner}}"
+    name  = "{{.github.name}}"
     pull_request {
-      branch = var.branch_regex
+      branch = "{{.branch_regex}}"
     }
   }
+  {{- else if index . "cloud_source_repository" -}}
```
MartinPetkov, line 167: You should mention somewhere that this is an option, either in the examples (maybe a new example?) or in the README. As it is, someone would have to read the .tf files to find out that cloud_source_repository is an option.

xingao267: Added docs in README.

I actually would like to hide this for now. For customers, we still would like to emphasize the importance of code review and running the two cloud build as presubmits before code submission. This is just for our internal testing. So probably not in a separate example to promote this.

MartinPetkov: Thanks. I understand, so I think it's okay for the examples to just use GitHub, but we shouldn't hide that we support CSR (it /is/ a Google product after all).

PR #228, opened 2020-05-29
Add script to check importer support for engine

MartinPetkov:
Don't run in CI yet, as it introduces toil to adding engine resources, and we haven't agreed yet that that's acceptable.


build/importer_engine_support_check.sh
```diff
@@ -0,0 +1,54 @@
+# Copyright 2020 Google LLC
```
umairidris, line 1: If we keep this script around I think it will be useful to convert this to Go at some point for maintainability.

MartinPetkov: What's the best practice for where to put a .go file like this? I don't want it to be a test because it shouldn't run every time, but it doesn't seem right to put it under cmd/.

umairidris: I am fine with adding a top level /scripts folder or under build/scripts.

https://github.com/golang-standards/project-layout/tree/master/scripts has some guidance.

MartinPetkov: Thanks. I'll move the file to go and put it under `scripts/`.

MartinPetkov: Done.

internal/tfengine/tfengine_test.go
```diff
@@ -54,7 +52,10 @@ func TestExamples(t *testing.T) {
 			path := filepath.Join(tmp, "live")
 
 			// Convert the configs not reference a GCS backend as the state bucket does not exist.
-			convertToLocalBackend(t, path)
+			err = ConvertToLocalBackend(path)
+			if err != nil {
```
umairidris, line 56: if err := ConvertToLocalBackend(); er != nil

also log func name and path in the t.Fatal log

internal/tfengine/tfengine.go
```diff
@@ -101,3 +102,41 @@ func dump(conf *Config, root, outputPath string) error {
...
 	return nil
 }
+
+// backendRE is a regex to capture GCS backend blocks in configs.
+// The 's' and 'U' flags allow capturing multi line backend blocks in a lazy manner.
+var backendRE = regexp.MustCompile(`(?sU)backend "gcs" {.*}`)
+
+// ConvertToLocalBackend converts all the Terraform backend blocks to "local".
+// path should be a path to the root of the configs generated by the Engine.
+func ConvertToLocalBackend(path string) error {
```
umairidris, line 112: Seems like this should go in a helper package. Maybe tfutil?

internal/tfengine/tfengine_test.go
```diff
@@ -52,9 +52,8 @@ func TestExamples(t *testing.T) {
 			path := filepath.Join(tmp, "live")
 
 			// Convert the configs not reference a GCS backend as the state bucket does not exist.
-			err = ConvertToLocalBackend(path)
-			if err != nil {
-				t.Fatalf("%v", err)
+			if err = ConvertToLocalBackend(path); err != nil {
```
umairidris, line 55: :=

Prefer to define vars in smaller scope

internal/tfengine/tfengine.go
MartinPetkov: I'm not sure. On the one hand it's specifically for the engine (to adjust its generated configs), on the other hand it would work on any .tf file. @xingao267 Any thoughts?

scripts/importer_engine_support_check.go
```diff
@@ -0,0 +1,121 @@
...
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/tfengine"
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/tfimport"
+)
+
+func main() {
```
xingao267, line 36: Add doc for what it checks for?

scripts/importer_engine_support_check.go
```diff
@@ -0,0 +1,121 @@
...
+	if err := tfengine.Run(full, tmp); err != nil {
+		log.Fatalf("tfengine.Run(%q, %q) = %v", full, tmp, err)
+	}
+
+	log.Printf("Converting all Terraform backend blocks to local")
+	path := filepath.Join(tmp, "live")
+	tfengine.ConvertToLocalBackend(path)
+
+	log.Printf("Initializing all modules in order to create and fill the .terraform/ dirs.")
+	plan := exec.Command("terragrunt", "plan-all")
```
xingao267, line 59: do you need to run plan-all to init all modules. Does init-all work?

scripts/importer_engine_support_check.go
```diff
@@ -0,0 +1,121 @@
...
+		_, supported := tfimport.Importers[r]
+		if unimportable || supported {
+			continue
+		}
+		unsupported = append(unsupported, r)
+	}
+
+	if len(unsupported) > 0 {
+		sort.Strings(unsupported)
+		log.Printf("Found unsupported resources:\n%v", strings.Join(unsupported, "\n"))
```
xingao267, line 84: early return in the `if` and remove the `else`

scripts/importer_engine_support_check.go
```diff
@@ -0,0 +1,121 @@
...
+	defer os.RemoveAll(tmp)
+
+	log.Printf("Generating configs from %v to %v", full, tmp)
+	if err := tfengine.Run(full, tmp); err != nil {
+		log.Fatalf("tfengine.Run(%q, %q) = %v", full, tmp, err)
+	}
+
+	log.Printf("Converting all Terraform backend blocks to local")
+	path := filepath.Join(tmp, "live")
+	tfengine.ConvertToLocalBackend(path)
```
umairidris, line 56: check err

scripts/importer_engine_support_check.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2020 Google LLC
```
umairidris, line 1: Because of how go packages work one "directory" becomes one binary, so this should be in a directory.

Also, maybe start this with a verb like

check_importer_matches_engine ?

internal/tfengine/tfengine.go
xingao267: Keeping it in the `tfengine` seems fine to me as it's pretty specific to the engine and currently not used by other tools.

umairidris: I don't have a strong opinion either way, just seeing tfengine.ConvertToLocalBackend seemed a little odd to me as it does not seem to be the engine's responsibility to be exporting this, nor is it actually used by the engine itself but rather only in the tests of the engine.

scripts/importer_engine_support_check.go
MartinPetkov: I left it as a single file because I saw this: https://github.com/hashicorp/terraform/blob/master/scripts/generate-plugins.go

I'll move it and rename it.

umairidris: Oh, I wonder if they run it like `go run generate-plugins.go`. In that case it may work even if you have multiple binaries in the same dir...

scripts/importer_engine_support_check.go
MartinPetkov: I have it higher up, above the package declaration. Should I put it above `main()` instead?

scripts/importer_engine_support_check.go
MartinPetkov: From what I can tell init-all doesn't work. It fails when it tries to init the root (I think?) and it says it initialized an empty directory.

scripts/importer_engine_support_check.go
MartinPetkov: They do run it like that, I think. I ran it like that when it was just one file. I don't mind multiple dirs.

internal/tfengine/tfengine.go
```diff
@@ -101,3 +102,41 @@ func dump(conf *Config, root, outputPath string) error {
...
+
+	// Replace all GCS backend blocks with local.
+	fn := func(path string, info os.FileInfo, err error) error {
+		if filepath.Ext(path) != ".tf" {
+			return nil
+		}
+
+		b, err := ioutil.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("ioutil.ReadFile(%q) = %v", path, err)
```
umairidris, line 126: Given this is no longer in the test, these logs should not read as a description e.g. read file %q: %v, path, err

scripts/check_importer_supports_engine/main.go
```diff
@@ -0,0 +1,123 @@
...
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/tfengine"
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/tfimport"
+)
+
+func main() {
+	full := "examples/tfengine/full.hcl"
```
umairidris, line 37: const outside main?

scripts/check_importer_supports_engine/main.go
```diff
@@ -0,0 +1,123 @@
...
+	if _, err := os.Stat(full); os.IsNotExist(err) {
```
umairidris, line 38: Make a single 'run' func that returns error, that way you don't need to do log.Fatals.

scripts/importer_engine_support_check.go
xingao267: Saw that now. No problem. It's good.

scripts/importer_engine_support_check.go
xingao267: Oh so I guess you are trying to actually use the plan from the console output. So then maybe update the `log.Printf("Initializing all modules in order to create and fill the .terraform/ dirs.")` which makes me think you just need to init the modules and use the contents in `.terraform` folder.

internal/tfengine/tfengine.go
MartinPetkov: I changed it, I think in the way that you meant.

scripts/importer_engine_support_check.go
MartinPetkov: You're actually right, that's exactly what I'm doing. I'd like to init and use the contents of the .terraform/ folders (multiple of them). As far as I can tell, `plan-all` does actually do what I would expect `init-all` to do, but it doesn't fail.

xingao267: Ok. Then consider using cmdRun() as you don't really use the output plan, and add a comment why using `plan-all` here.

internal/tfengine/tfengine.go
```diff
@@ -101,3 +102,41 @@ func dump(conf *Config, root, outputPath string) error {
...
+		b = backendRE.ReplaceAll(b, nil)
+		if err := ioutil.WriteFile(path, b, 0644); err != nil {
+			return fmt.Errorf("write file %q: %v", path, err)
+		}
+
+		return nil
+	}
+
+	if err := filepath.Walk(path, fn); err != nil {
+		return fmt.Errorf("walk files: %v", err)
```
umairidris, line 138: walk %q: %v, path, err

internal/tfengine/tfengine.go
```diff
@@ -101,3 +102,41 @@ func dump(conf *Config, root, outputPath string) error {
...
+// backendRE is a regex to capture GCS backend blocks in configs.
+// The 's' and 'U' flags allow capturing multi line backend blocks in a lazy manner.
+var backendRE = regexp.MustCompile(`(?sU)backend "gcs" {.*}`)
+
+// ConvertToLocalBackend converts all the Terraform backend blocks to "local".
+// path should be a path to the root of the configs generated by the Engine.
+func ConvertToLocalBackend(path string) error {
+	// Overwrite root terragrunt file with empty file so it doesn't try to setup remote backend.
+	if err := ioutil.WriteFile(filepath.Join(path, "terragrunt.hcl"), nil, 0664); err != nil {
+		return fmt.Errorf("ioutil.Write terragrunt root file: %v", err)
```
umairidris, line 115: write Terragrunt root at %q: %v, path, err

PR #235, opened 2020-06-01
Add whitelist capability to deletion check

### LINKED ISSUE #101: Support blocking on deletion for specific resources
It would be great if we can have a whitelist or blacklist mode for resources and make the deletion check block on it.

Examples:
- in mode whitelist deletions of specific types are allowed. If a new resource wants to be deleted and allowed then first send a PR allowing that resource or resource type.
- in mode blacklist deletions of specific types are not allowed. Thus for example you can prevent any deletion of sensitive resources like GKE cluster or PHI holding resources like healthcare stores.

MartinPetkov:
I think this is a good balance between a simple implementation and usefulness.

Fixes #101


templates/tfengine/components/cicd/manual/README.md
```diff
@@ -95,6 +97,21 @@ Every new push to the Pull Request at the configured branches automatically
 triggers presubmit runs. To manually re-trigger CI jobs, comment `/gcbrun` in the
 Pull Ruquest.
 
+The deletion check optionally accepts a whitelist of resources to ignore. To
```
xingao267, line 100: Add more doc around in which format should users input the resource. Is it a terraform state address or resource name or something.

templates/tfengine/components/cicd/manual/README.md
```diff
@@ -95,6 +97,31 @@ Every new push to the Pull Request at the configured branches automatically
...
+
+```text
+networks
+^module.cloudsql.module.safer_mysql.google_sql_database.default$
+google_sql_user.db_users\["user-creds"\]
+```
+
+Each line whitelists, respectively:
+
+1. Any resource whose address contains the string "network".
```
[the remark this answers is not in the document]
MartinPetkov: You're right, I'll fix it.

templates/tfengine/components/cicd/manual/configs/tf-plan.yaml
```diff
@@ -42,14 +42,14 @@ steps:
...
-    args: ["../cicd/configs/terragrunt.sh", "plan-all"]
+    args: ["../cicd/configs/terragrunt.sh", "plan-all", "-out='plan.tfplan'"]
     dir: "${_TERRAFORM_ROOT}/live"
     id: Speculative plan (live)
 
   # Check for delete operations as an FYI, it won't fail the build.
   - name: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:b007041726d4f2d7e5270b8fcfc47656850242186825611b33ee97b3867d74ed"
     entrypoint: bash
-    args: ["../cicd/configs/tf-deletion-check.sh"]
+    args: ["../cicd/configs/tf-deletion-check.sh", "../cicd/configs/tf-delete-whitelist.txt"]
```
xingao267, line 52: we can rename `tf-delete-whitelist.txt` to tf-deletion-whitelist.txt` to be consistent with `tf-deletion-check.sh`

MartinPetkov: Done.

templates/tfengine/components/cicd/manual/configs/tf-plan.yaml
```diff
@@ -42,14 +42,14 @@ steps:
 
   - name: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:b007041726d4f2d7e5270b8fcfc47656850242186825611b33ee97b3867d74ed"
     entrypoint: bash
-    args: ["../cicd/configs/terragrunt.sh", "plan-all"]
+    args: ["../cicd/configs/terragrunt.sh", "plan-all", "-out='plan.tfplan'"]
```
xingao267, line 45: terragrunt.sh's plan-all actually includes -out=plan.tfplan as a flag, doesn't that work for you?

MartinPetkov: Oh, you're right, I didn't notice this was running a .sh script and not terragrunt directly. Thank you, I'll remove this.

PR #237, opened 2020-06-01
policygen: support allow roles and bindings policy

xingao267:
#152


internal/policygen/iam.go
```diff
@@ -85,3 +126,26 @@ func validate(instance map[string]interface{}, mandatoryFields []string) error {
...
+	out, err := rn.CmdOutput(cmd)
+	if err != nil {
+		return "", fmt.Errorf("failed to get project number for project %q: %v", id, err)
+	}
+
+	type project struct {
+		ProjectNumber string `json:"projectNumber"`
+	}
+
+	var p project
```
MartinPetkov, line 141: Nit: Can do `var p struct { ... }`

internal/policygen/iam.go
```diff
@@ -85,3 +126,26 @@ func validate(instance map[string]interface{}, mandatoryFields []string) error {
 	}
 	return nil
 }
+
+func projectNumber(rn runner.Runner, id string) (string, error) {
+	cmd := exec.Command("gcloud", "projects", "describe", id, "--format", "json")
```
MartinPetkov, line 131: You can actually replace this with
```go
cmd := exec.Command("gcloud", "projects", "describe", id, "--format='value(projectNumber)'")
```

internal/policygen/iam.go
```diff
@@ -15,62 +15,103 @@
...
+				bindings[key] = make(roleBindings)
 			}
 
-			bindings = append(bindings, &binding{
-				Type:   t,
-				ID:     ins[idField].(string), // Type checked in validate()
-				Role:   ins["role"].(string),
-				Member: ins["member"].(string),
-			})
+			bindings[key][ins["role"].(string)] = append(bindings[key][ins["role"].(string)], ins["member"].(string))
```
MartinPetkov, line 111: `ins["role"].(string)` is repeated, should be extracted. Or just use `id`.

internal/policygen/iam.go
```diff
@@ -15,62 +15,103 @@
...
 }
 
-func generateIAMBindingsPolicies(resources []*states.Resource, outputPath string) error {
+// All IAM members associated with a single role.
+type roleBindings map[string][]string
 
-	bindings, err := bindings(resources)
-	if err != nil {
+// All roles associated with a root resource (organization, folder or project)
+var bindings = make(map[root]roleBindings)
```
MartinPetkov, line 39: I find it smelly that this is a global variable (sort of) and is being secretly modified by `allBindings`. Why not have `allBindings` generate and return it?

xingao267: Done

internal/policygen/iam.go
xingao267: Done

internal/policygen/iam.go
xingao267: `p` is actually to be passed to the `Unmarshal` function to hold the struct parsed from `out`.

internal/policygen/policygen_test.go
```diff
@@ -17,10 +17,40 @@ package policygen
...
 )
 
+type Fake struct{}
+
+func (*Fake) CmdRun(*exec.Cmd) error {
+	return nil
+}
+
+func (*Fake) CmdOutput(cmd *exec.Cmd) ([]byte, error) {
+	contains := func(s string, subs ...string) bool {
```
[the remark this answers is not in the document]
xingao267: It's used at L43 and not used anywhere else, so just made this a inner function of `CmdOutput`

internal/policygen/iam.go
xingao267: I actually debated a little between the two but decided to choose json format as I found it's more structured and somewhat feels a bit more "reliable". Another reason is I might need other fields from the `projects describe` output, so the full `json` output can be more robust in that case.

internal/policygen/policygen_test.go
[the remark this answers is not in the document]
MartinPetkov: Right, but then the function gets redefined every time CmdOutput is called, but there's no reason to. I think it's fine if it's just a non-exported function in this same file.

internal/policygen/iam.go
MartinPetkov: I see. I would still argue for just using `value()`; you don't need other fields yet, and it would simplify this function a lot, since it's only returning the `projectNumber`. It can more or less become just 
```
func projectNumber(...) (...) {
    return rn.CmdOutput(exec.Command("gcloud", "projects", "describe", id, "--format='value(projectNumber)'")
}
```

(depending on if you want to check that it's not empty)

internal/policygen/policygen_test.go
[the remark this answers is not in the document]
xingao267: Done

internal/policygen/iam.go
xingao267: Changed. Although it's not that simply as there is still post-processing needs to be done.

templates/policygen/forseti/tf_based/iam_allow_bindings.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+apiVersion: constraints.gatekeeper.sh/v1alpha1
+kind: GCPIAMAllowedBindingsConstraintV1
+metadata:
+  name: whitelist_owner_role
+spec:
+  severity: high
+  match:
+    target:
+    - "{{get . "target"}}"
```
umairidris, line 23: This will put the value nil i think or no type if there is no get value. Is that ok? If it should be required then this should be {{.target}}

templates/policygen/forseti/tf_based/iam_allow_bindings.yaml
```diff
@@ -0,0 +1,30 @@
...
+  parameters:
+    mode: whitelist
+    role: {{get . "role"}}
+    members:
+    {{- range .bindings}}
```
umairidris, line 28: Why is this field called bindings when it's for members?

internal/policygen/policygen_test.go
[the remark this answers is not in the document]
umairidris: While either is fine I would lean towards Martin's suggestion here. I only define funcs in line if they're very small, or close over other vars in the outer func. Defining this in the top level as a private func is fine here (though doesn't matter cus its a test anyway).

internal/policygen/iam.go
```diff
@@ -85,3 +128,18 @@ func validate(instance map[string]interface{}, mandatoryFields []string) error {
 	}
 	return nil
 }
+
+func projectNumber(rn runner.Runner, id string) (string, error) {
+	cmd := exec.Command("gcloud", "projects", "describe", id, "--format=value(projectNumber)")
```
umairidris, line 133: This is fine as well, but my preference is for JSON because that will be consistent with any gcloud lookup (e.g. if you want to get a list of values, etc).

umairidris: Have we documented gcloud as a dep?

xingao267: Added to README

templates/policygen/forseti/tf_based/iam_allow_bindings.yaml
xingao267: Done

templates/policygen/forseti/tf_based/iam_allow_bindings.yaml
xingao267: Done

internal/policygen/iam.go
xingao267: Done

PR #238, opened 2020-06-02
Add support for secrets

### LINKED ISSUE #187: Support secrets
Both resources and data

```
resource "google_secret_manager_secret" "secret" {
  provider = google-beta

  secret_id = "foo"
  project   = var.project_id

  replication {
    automatic = true
  }
}

resource "google_secret_manager_secret_version" "value" {
  secret      = google_secret_manager_secret.foo.id
  secret_data = random_password.foo.result
}

data "google_secret_manager_secret_version" "secret" {
  provider = google-beta
  project  = var.secrets_project_id
  secret   = "foo"
}
```

umairidris:
Also:

- In addition to terragrunt leaf, add a terraform leaf recipe as well. This will be used for deployments such as secrets deployment that are not handled by Terragrunt
- Remove 'disabled' block. Simply comment out blocks and put a TODO to uncomment them. This is simpler and more intuitive and works with arbitrary configs.

Fix #187


examples/tfengine/full.hcl
```diff
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{$base := "../../templates/tfengine/recipes"}}
+# {{$base := "../../templates/tfengine/recipes"}}
```
xingao267, line 15: did you intend to comment out this line?

examples/tfengine/simple.hcl
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{$base := "../../templates/tfengine/recipes"}}
+# {{$base := "../../templates/tfengine/recipes"}}
```
xingao267, line 15: This one as well.

templates/tfengine/components/cicd/manual/main.tf
```diff
@@ -68,6 +68,7 @@ locals {
   cloudbuild_devops_roles = [
     # Enable Cloud Build SA to list and enable APIs in the devops project.
     "roles/serviceusage.serviceUsageAdmin",
+    "roles/secretmanager.viewer",
```
xingao267, line 71: You might need admin permission. I got an error with just viewer permission with a full clean deployment. But it also depends on if CICD is actually filling values for some secrets.

examples/tfengine/full.hcl
umairidris: Yeah, else I cant run terragrunt hclfmt

templates/tfengine/components/cicd/manual/main.tf
MartinPetkov: Or at least `roles/secretmanager.secretAccessor`.

templates/tfengine/components/cicd/manual/main.tf
```diff
@@ -54,12 +54,15 @@ locals {
   cloudbuild_sa_viewer_roles = [
     "roles/browser",
     "roles/iam.securityReviewer",
+    "roles/secretmanager.viewer",
```
xingao267, line 57: why do we need these additional permissions?

umairidris: Doesn't the CICD need these permissions to access / deploy secrets?

xingao267: But the SA already has owner role to each project though. And we don't want users to put secrets in the devops project, so shouldn't grant SA secret admin role in devops project.

umairidris: Oh right.. good point. I'll remove this.

docs/tfengine/secrets.md
```diff
@@ -0,0 +1,48 @@
+# Secrets
+
+The secrets deployment is used to define sensitive values that should not be
+set in plaintext. Secrets are deployed in the devops project.
```
MartinPetkov, line 4: >Secrets are deployed in the devops project
Is that still true? Looks like secrets get their own "example-secrets" project now.

examples/tfengine/full.hcl
```diff
@@ -107,6 +85,42 @@ template "foundation" {
...
+      ]
+    }
+    resources = {
+      secrets = [
+        {
+          secret_id = "manual-sql-db-user"
+        },
+        {
+          secret_id   = "auto-sql-db-password"
+          secret_data = "$${random_password.db.result}" // Use $$ to escape.
```
MartinPetkov, line 109: To escape what?

MartinPetkov, line 109: I feel like if a secret is "auto-*", it would be neat if the engine generated a "random_password" block for it. Would that be hard to add?

templates/tfengine/components/terraform/main/main.tf
```diff
@@ -18,7 +18,12 @@ terraform {
     google      = "~> 3.0"
     google-beta = "~> 3.0"
   }
-  backend "gcs" {}
+  backend "gcs" {
+    {{- if not (get . "disable_gcs_backend_config")}}
```
MartinPetkov, line 22: "not get disabled" is a double negative. Why not replace with "if get . "enable_..."?

templates/tfengine/recipes/org/project.hcl
```diff
@@ -38,7 +33,7 @@ template "project" {
 {{if has . "resources"}}
 template "resources" {
   recipe_path = "../project/resources.hcl"
-  output_path = "./{{.project.project_id}}/resources"
+  output_path = "././resources"
   flatten {
```
MartinPetkov, line 37: ././ should just be ./

cmd/tfengine/README.md
```diff
@@ -119,7 +119,7 @@ terraform init
 terraform plan
 terraform apply
 
-# Step 6 (Optional): Deploy Continuous Integration (CI) and Continuous
+# Step 7 (Optional): Deploy Continuous Integration (CI) and Continuous
```
xingao267, line 122: seems should still be Step 6?

umairidris: I forgot to add the new step 6 :D

templates/tfengine/components/bootstrap/main.tf
```diff
@@ -48,6 +48,7 @@ module "project" {
   skip_gcloud_download    = true
   activate_apis = [
     "cloudbuild.googleapis.com",
+    "secretmanager.googleapis.com",
```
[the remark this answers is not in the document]
umairidris: Reverted

templates/tfengine/recipes/org/project.hcl
umairidris: Fixed

templates/tfengine/components/terraform/main/main.tf
umairidris: Then all users will always need to pass in enable_gcs_backend. In this case, if no field is passed then this block will be enabled.

We had an enabled helper before but given this field is only set by internal templates I don't think double negative is that big a deal. External users aren't expected to be passing this field.

examples/tfengine/full.hcl
umairidris: That's a good idea, let me put a TODO for this.

umairidris: updated

docs/tfengine/secrets.md
umairidris: Done.

templates/tfengine/recipes/org/org_policies.hcl
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 template "terragrunt" {
-  component_path = "../../components/terragrunt/leaf"
+  recipe_path = "../terraform/terragrunt.hcl"
```
xingao267, line 16: why does it become a recipe?

cmd/tfengine/README.md
xingao267: Since the secrets are in live/ now. Maybe move it after CICD.

templates/tfengine/recipes/org/org_policies.hcl
umairidris: Terragrunt recipe no longer sets up the terraform configs, only terragrunt.hcl

cmd/tfengine/README.md
umairidris: Agreed

PR #246, opened 2020-06-03
policygen: update README

xingao267:


cmd/policygen/README.md
```diff
@@ -65,8 +65,16 @@ terraform apply
...
 #### Policy Library Constraints
 
-To use Policy Library Constraints with **Forseti**, follow
-[How to use Forseti Config Validator](https://github.com/forseti-security/policy-library/blob/master/docs/user_guide.md#how-to-use-forseti-config-validator).
+* To use Policy Library Constraints with **Forseti**, follow
+    [How to use Forseti Config Validator](https://github.com/forseti-security/policy-library/blob/master/docs/user_guide.md#how-to-use-forseti-config-validator).
 
-To use Policy Library Constraints with **Terraform Validator**, follow
-[How to use Terraform Validator](https://github.com/forseti-security/policy-library/blob/master/docs/user_guide.md#how-to-use-terraform-validator).
+    The `target` values under `match` block in the auto-generated policies might
```
MartinPetkov, line 71: Re "might", could you expand on when they need to be adjusted?

xingao267: done, ptal

MartinPetkov: Makes a lot more sense now, but it's definitely surprising. I would expect target to work without specifying the parents too.

xingao267: yea, it's a bit tricky there. I'm also trying to work with Forseti owners to make this easier. a glob feature is supposed to work but seems not. Filed https://github.com/forseti-security/config-validator/issues/142, so hopefully later this can be made easy.

PR #250, opened 2020-06-03
Importer: Add google_resource_manager_lien

MartinPetkov:


internal/tfimport/importer/google_resource_manager_lien.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/terraform"
+)
+
+// ResourceManagerLien defines a struct with the necessary information for a google_resource_manager_lien to be imported.
+type ResourceManagerLien struct{}
+
+// ImportID returns the ID of the resource to use in importing.
+func (i *ResourceManagerLien) ImportID(rc terraform.ResourceChange, pcv ConfigMap, interactive bool) (string, error) {
+	// Get required fields.
+	fieldsMap, err := loadFields([]string{"parent", "name"}, interactive, rc.Change.After, pcv)
```
umairidris, line 31: The name is not known from beforehand from what I remember. 

`name - A system-generated unique identifier for this Lien.`

So how would loadFields know the value?

In V2 I used to list all liens and match against one with the same set of restrictions.

internal/tfimport/tfimport_test.go
```diff
@@ -788,6 +788,14 @@ func TestImportersSanity(t *testing.T) {
 			},
 			"projects/my-project/topics/my-topic",
 		},
+		{
+			"google_resource_manager_lien",
+			map[string]interface{}{
+				"parent": "projects/1234567890",
+				"name":   "auto-generated-name",
```
umairidris, line 795: This name wouldnt be in the plan

internal/tfimport/importer/google_resource_manager_lien.go
MartinPetkov: I originally had it like this because loadFields will ask the user for it. I assumed the user would look it up themselves.

I've now modified the PR to fetch the liens remotely and present them as choices.

umairidris:
Does it present choices if there is only one lien and it matches the restrictions? I don't think we need to in that case and can just import it.

MartinPetkov:
> Does it present choices if there is only one lien and it matches the restrictions? I don't think we need to in that case and can just import it.

It doesn't. Even if there's only one choice, it will present it.

I don't think it's safe to auto-import if there's only one. It could still be some other lien added by someone else or by another system. For example, when testing, I had another lien added by the Project Factory CFT module, and it would be wrong to import that into the brand new one I just added for testing.

Issue #251, opened 2020-06-04
Check in generated configs into github as well


umairidris:

umairidris:
#264

PR #253, opened 2020-06-04
check in policies generated from examples

xingao267:
#251


MartinPetkov:
Maybe add a check in CI to make sure the generated files stay up to date.

MartinPetkov:
> Maybe add a check in CI to make sure the generated files stay up to date.

I opened #254

PR #255, opened 2020-06-04
Add CI check for generated files

### LINKED ISSUE #254: Check that policygen generated files are up to date in CI
Since https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/253, generated files for policygen are checked in with the repo. We should add a check in CI to make sure they stay up-to-date.

MartinPetkov:
Fixes #254

Will fail until https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/253 is checked in


build/presubmit.yaml
```diff
@@ -28,5 +28,10 @@ steps:
   args: ["build/go_check.sh"]
   id: Go code check
 
+- name: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:b007041726d4f2d7e5270b8fcfc47656850242186825611b33ee97b3867d74ed"
+  entrypoint: "sh"
+  args: ["build/gen_check.sh"]
+  id: Generated files check
+
```
xingao267, line 35: we have more tests now and we can add the `waitFor: ["-"]` to each step so they run in parallel

build/gen_check.sh
```diff
@@ -0,0 +1,49 @@
...
+  git init
+  git config user.email "noname@nomail.com"
+  git config user.name "No Name"
+  git add .
+  git commit -a --allow-empty-message -m ''
+  set +x
+fi
+
+# Generate files and check for changes.
+cmd='go run ./cmd/policygen --config_path=examples/policygen/config.yaml --state_path=examples/policygen/example.tfstate --output_path=examples/policygen/generated'
```
xingao267, line 37: Can we just generate to a tmp directory, e.g. /tmp/policygen, then `diff -r /tmp/poligen examples/policygen/generated` ?

build/presubmit.yaml
MartinPetkov: I'll open a separate CL.

build/gen_check.sh
MartinPetkov: That's a great idea. Updated.

build/presubmit.yaml
MartinPetkov: Opened https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/263

xingao267:
Thanks for adding this!

Issue #156, opened 2020-05-20
Format and add license at the end of policygen execution.


xingao267:

xingao267:
The license part is done.

The format part is duplicated in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/240

PR #258, opened 2020-06-05
add helper function to add license in generated files

xingao267:
#156


internal/licenseutil/licenseutil.go
```diff
@@ -0,0 +1,101 @@
...
+
+`
+
+// AddLicense adds Google LLC Apache 2.0 license header to Terraform files in dir.
+func AddLicense(dir string) error {
+	fs, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return fmt.Errorf("read dir %q: %v", dir, err)
+	}
+	if len(fs) == 0 {
```
MartinPetkov, line 49: I don't think this should be an error. The function would just return, having done nothing.

internal/licenseutil/licenseutil.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("found no files in %q to write", dir)
+	}
+
+	for _, f := range fs {
```
MartinPetkov, line 53: Why not use `filepath.Walk()`? I do that in `scripts/check_importer_supports_engine/main.go`.

internal/licenseutil/licenseutil.go
```diff
@@ -0,0 +1,101 @@
...
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
+	}
+
+	if bytes.Contains(b, []byte("Copyright 2020 Google LLC")) {
+		return nil
+	}
+
+	b = append([]byte(licenseString), b...)
```
MartinPetkov, line 85: If it's always getting converted to a bytes array, you might as well just make it that type instead of `string`.

internal/licenseutil/licenseutil.go
```diff
@@ -0,0 +1,101 @@
...
+	return ioutil.WriteFile(path, b, fmode)
+}
+
+func isTerraformFile(name string) bool {
+	name = strings.ToLower(filepath.Base(name))
+	if name == "terragrunt.hcl" {
+		return true
+	}
+
+	ext := name
```
MartinPetkov, line 95: I'm not sure what this is doing. Could the basename of a file ever be just ".tf" or ".tfvars"?

It seems like if `filepath.Ext()` returns `""`, this function should just return `false`.

internal/licenseutil/licenseutil.go
xingao267: Done

PR #265, opened 2020-06-08
Add support for bastion hosts and compute routers & nats

umairidris:
Also add new helper funcmap `hcl`


examples/tfengine/generated/full/live/prod/team1/example-prod-apps/project/terragrunt.hcl
```diff
@@ -29,5 +30,6 @@ dependency "networks" {
 }
 
 inputs = {
-  folder_id = dependency.parent_folder.outputs.name
+  folder_id = "${dependency.parent_folder.outputs.name}"
```
MartinPetkov, line 33: Is this correct? I've previously seen it be like it was before, without the `${}`.

templates/tfengine/components/project/bastion_hosts/main.tf
```diff
@@ -0,0 +1,43 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */ -}}
+
+{{range .bastion_hosts -}}
+module "{{resourceName .name}}" {
+  source = "git::https://github.com/terraform-google-modules/terraform-google-bastion-host.git?ref=umairidris-patch-1"
```
MartinPetkov, line 17: Leave a TODO this this should change to an official release, once available.

examples/tfengine/generated/full/live/prod/team1/example-prod-apps/project/terragrunt.hcl
umairidris: I changed the template to use the new `hcl` func in funcmap, that's why this is now needed. Before its value was just set directly.

MartinPetkov: I see. How come terragrunt can read this kind of syntax though, in the generated files?

umairidris: They added support for it.

https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L58
https://github.com/gruntwork-io/terragrunt/blob/master/config/config.go#L349

templates/tfengine/components/project/bastion_hosts/main.tf
umairidris: Done.

examples/tfengine/generated/full/live/prod/team1/example-prod-apps/project/terragrunt.hcl
MartinPetkov: I see.

examples/tfengine/full.hcl
```diff
@@ -281,7 +334,8 @@ data "google_secret_manager_secret_version" "db_password" {
   project = "example-secrets"
 }
 EOF
-      } */
+ */
```
xingao267, line 337: make this align with the /* TODO(user) ?

umairidris: This is done by the terragrunt hclfmt

PR #283, opened 2020-06-10
support per-recipe schema

umairidris:
#74

### DIFF
--- a/internal/tfengine/config.go
+++ b/internal/tfengine/config.go
@@ -129,6 +129,8 @@ func loadConfig(path string, data map[string]interface{}) (*Config, error) {
 }
 
 func hclToJSON(b []byte) ([]byte, error) {
+	// Directly trying to unmarshal to cty.Value doesn't seem to work,
+	// so wrap in a dummy field.
 	wrap := struct {
 		Value *cty.Value `hcl:"value"`
 	}{}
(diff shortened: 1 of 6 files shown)


internal/tfengine/config.go
```diff
@@ -61,46 +64,35 @@ func (c *Config) Init() error {
 		}
 	}
 
+	if c.SchemaCty != nil {
+		c.Schema, err = ctyValueToMap((c.SchemaCty))
```
MartinPetkov, line 68: Typo: Extra set of `()`

internal/tfengine/config.go
```diff
@@ -135,3 +127,39 @@ func loadConfig(path string, data map[string]interface{}) (*Config, error) {
 	}
 	return c, nil
 }
+
+func hclToJSON(b []byte) ([]byte, error) {
+	// Directly trying to unmarshal to cty.Value doesn't seem to work,
+	// so wrap in a dummy field.
+	wrap := struct {
```
MartinPetkov, line 134: Lower down, you use the construct `var jr struct { ... }`, while here you do what (I think) is the same thing but with `wrap := struct { ... }{}`.

Should choose one and use it for consistency. Personally I slightly prefer `var` over `:=`.

PR #285, opened 2020-06-11
Support overriding Terraform resource names

### LINKED ISSUE #203: Support overriding resource names
There are often where our resourceName helper does not pick a suitable or best possible resource name. In this case we should allow the resource name to be explicitly overridden.

umairidris:
Fix #203 

This is used in the case we don't create an appropriate terraform resource name, a user can override it.

### DIFF
--- a/internal/template/funcmap.go
+++ b/internal/template/funcmap.go
@@ -101,7 +101,7 @@ func resourceName(m map[string]interface{}, k string) (string, error) {
 
 	name, ok := v.(string)
 	if !ok {
-		return "", fmt.Errorf("resource_name value %v is not a string", v)
+		return "", fmt.Errorf("resource name value %v is not a string", v)
 	}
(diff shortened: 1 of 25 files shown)


internal/template/funcmap.go
```diff
@@ -90,6 +90,18 @@ func hclField(m map[string]interface{}, key string, req bool) (string, error) {
 
 // resourceName builds a Terraform resource name.
 // GCP resource names often use "-" but Terraform resource names should use "_".
-func resourceName(s string) string {
-	return strings.Replace(s, "-", "_", -1)
+func resourceName(m map[string]interface{}, k string) (string, error) {
```
MartinPetkov, line 93: Maybe update the function comment now that it does more.

umairidris: Good point, done

PR #289, opened 2020-06-12
More schemas for basic recipes

umairidris:
#74

### DIFF
--- a/templates/tfengine/recipes/org/audit.hcl
+++ b/templates/tfengine/recipes/org/audit.hcl
@@ -12,6 +12,48 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+schema = {
+  title = "Org Audit Recipe"
+  required = [
+    "project_id",
+    "dataset_name",
+    "bucket_name",
+    "auditors",
+    "bigquery_location",
+    "storage_location",
+  ]
+  properties = {
+    project_id = {
+      description = "ID of the project to host audit resources."
+      type        = "string"
+    }
+    dataset_name = {
+      description = "Name of the Bigquery Dataset to store 1 year audit logs."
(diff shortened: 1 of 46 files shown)


templates/tfengine/recipes/org/audit.hcl
```diff
@@ -12,6 +12,48 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+schema = {
```
MartinPetkov, line 15: Seeing it now, I feel that having this schema here like this really helps!

templates/tfengine/recipes/org/audit.hcl
```diff
@@ -12,6 +12,48 @@
...
+      type        = "string"
+    }
+    bucket_name = {
+      description = "Name of GCS Bucket to store 7 year audit logs."
+      type        = "string"
+    }
+    auditors = {
+      description = <<EOF
+        This group will be granted viewer access to the audit log dataset and
+        bucket as well as security reviewer permission on the entire org.
```
xingao267, line 41: on the entire org --> on the root resource specified

templates/tfengine/recipes/org/monitor.hcl
```diff
@@ -12,6 +12,36 @@
...
+    project_id = {
+      description = "ID of the project to host monitoring resources."
+      type        = "string"
+    }
+    domain = {
+      description = "Domain for the Forseti instance."
+      type        = "string"
+    }
+    cloud_sql_region = {
+      description = "Default location for compute instances."
```
xingao267: for Cloud SQL

xingao267: nvm. This is fixed in a new commit.

PR #293, opened 2020-06-12
Merge engine and policygen schema logic for org policies

### LINKED ISSUE #174: Invoke policygen functions to generate org policies in tfengine

xingao267:
Fixes #174 

The templating logic that read config and output to appropriate locations will stay in engine or policygen. The most important part is to avoid duplicating the schema.

### DIFF
--- a/internal/policygen/load.go
+++ b/internal/policygen/load.go
@@ -15,6 +15,7 @@
 package policygen
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


internal/policygen/load.go
```diff
@@ -38,6 +39,23 @@ type config struct {
 	Schema    map[string]interface{} `json:"schema,omitempty"`
 }
 
+func ValidateOrgPoliciesConfig(conf map[string]interface{}, allowAdditionalProperties bool) error {
+	if !allowAdditionalProperties {
+		orgPoliciesSchema = append(orgPoliciesSchema, []byte("additionalProperties = false")...)
```
MartinPetkov, line 44: It seems to me that this will modify the global variable and will always have `additionalProperties = false` for every subsequent function call. Maybe this should be `:=`, to create a locally scoped variable?

internal/policygen/load.go
```diff
@@ -49,7 +67,7 @@ func loadConfig(path string) (*config, error) {
 		return nil, fmt.Errorf("convert config to JSON: %v", err)
 	}
 
-	sj, err := hcl.ToJSON([]byte(schema))
+	sj, err := hcl.ToJSON(schema)
```
MartinPetkov, line 70: Where is the `sj` variable used? I can't see it in this function after it's assigned.

xingao267: It's used at line 75

internal/policygen/load.go
xingao267: Done

internal/policygen/schema.go
```diff
@@ -16,7 +16,7 @@ package policygen
 
 // TODO(https://github.com/golang/go/issues/35950): Move this to its own file.
 
-const schema = `
+var schema = []byte(`
```
umairidris, line 19: Does this need to be changed? You're not modifying this var it seems

xingao267: It seems I can't make a byte slice an const

xingao267: and then the byte array is directly used in the hcl.ToJSON func

PR #294, opened 2020-06-13
support cscc notification

xingao267:


templates/tfengine/components/org/monitor/forseti/main.tf
```diff
@@ -88,4 +88,11 @@ module "forseti" {
 
   manage_rules_enabled     = false
   config_validator_enabled = true
+
+  {{- if has . "cscc_source_id"}}
+
+  # Enable Cloud Security Command Center (CSCC) notification.
+  cscc_violations_enabled  = true
+  cscc_source_id           = "{{.cscc_source_id}}"
```
MartinPetkov, line 96: I think you can use the `hclField` func for both of these fields.

xingao267: Discussed offline, will leave as it is right now.

PR #298, opened 2020-06-15
Add devops recipe, remove foundation

### LINKED ISSUE #173: Allow cicd field to be set under devops
Currently the cicd field is a sibling to devops. This requires duplicationg the devops project id and state bucket.

We should optionally allow the cicd to be set as a child of the devops field so that the common fields don't need to be duplicated.

umairidris:
Fix #173

#74


examples/tfengine/full.hcl
```diff
@@ -89,6 +60,41 @@ template "foundation" {
         "secretmanager.googleapis.com",
       ]
     }
+
+    # TODO(user): Uncomment and re-run the engine after generated bootstrap module has been deployed.
+    # Run `terraform init` in the bootstrap module to backup its state to GCS.
+    # bootstrap_gcs_backend = true
```
xingao267, line 66: move this to the beginning of devops block?

internal/policygen/schema.go
```diff
@@ -65,7 +65,6 @@ var orgPoliciesSchema = []byte(`
 title = "Google Cloud Organization Policies Config Schema"
 required = [
   "parent_type",
-  "parent_id",
```
xingao267, line 68: project_id should still be supported

templates/policygen/org_policies/main.tf
```diff
@@ -17,14 +17,20 @@ limitations under the License. */ -}}
 # See the following resources for the details of policies enforced.
 
 {{- $type_field := printf "%s_id" .parent_type}}
+{{- $parent_id := ""}}
+{{- if eq .parent_type "organization"}}
+{{- $parent_id = .org_id}}
+{{- else}}
```
xingao267, line 23: `else if eq .parent_type "folder"`

and leave the `else` for project

templates/tfengine/recipes/org/devops.hcl
```diff
@@ -0,0 +1,84 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+schema = {
+  title = "Devops Recipe"
+  required = [
+    "project_id",
+    "org_id",
```
xingao267, line 19: org_id shouldn't be required as folder_id can be specified instead of org_id

Why are we changing parent_id to org_id + folder_id?

umairidris: The reason was parent_id was just duplicating the org_id/folder_id, etc that was already set in the config.

internal/policygen/schema.go
xingao267, line 68: This should be reverted as well

templates/tfengine/components/cicd/manual/main.tf
```diff
@@ -126,9 +126,9 @@ resource "google_storage_bucket_iam_member" "cloudbuild_state_iam" {
 resource "google_{{.parent_type}}_iam_member" "cloudbuild_sa_{{.parent_type}}_iam" {
   for_each = toset(var.continuous_deployment_enabled ? local.cloudbuild_sa_editor_roles : local.cloudbuild_sa_viewer_roles)
   {{- if eq (get . "parent_type") "organization"}}
-  org_id   = {{.parent_id}}
```
xingao267, line 129: this as well

templates/tfengine/recipes/org/devops.hcl
```diff
@@ -0,0 +1,84 @@
...
+    "billing_account",
+    "org_admin",
+    "project_owners",
+  ]
+  properties = {
+    project_id = {
+      description = "Project ID to host devops related resources such as state bucket and CICD."
+      type        = "string"
+    }
+    org_id = {
```
xingao267, line 29: should this be removed now?

PR #301, opened 2020-06-16
Add Sprig

MartinPetkov:
Docs: https://github.com/Masterminds/sprig

This adds a lot of helper functions that make it easier to write templates.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,9 @@ go 1.14
 
 require (
 	cloud.google.com/go v0.58.0 // indirect
+	github.com/Masterminds/goutils v1.1.0 // indirect
+	github.com/Masterminds/semver v1.5.0 // indirect
+	github.com/Masterminds/sprig v2.22.0+incompatible
 	github.com/agext/levenshtein v1.2.3 // indirect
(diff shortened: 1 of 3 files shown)


umairidris:
I tried supporting this initially and decided against doing this for a couple of reasons:

- a lot of deps (hard to import back to g3)
- we have custom logic in a lot of our funcs (e.g. supporting multiple levels)

the most important funcs like getting default value, checking if key is there, marshalling to hcl etc are already there, so I don't think we need all the funcs of sprig. It will add further confusion on when to use what.

MartinPetkov:
> I tried supporting this initially and decided against doing this for a couple of reasons:
> 
> * a lot of deps (hard to import back to g3)
> * we have custom logic in a lot of our funcs (e.g. supporting multiple levels)
> 
> the most important funcs like getting default value, checking if key is there, marshalling to hcl etc are already there, so I don't think we need all the funcs of sprig. It will add further confusion on when to use what.

I see. Truthfully I only need it for the `replace` function right now, and I thought I might as well add the whole thing since I've found it really useful in the past.

I'll close this PR and just add `replace` for now.

PR #305, opened 2020-06-16
Add support for top-level data block recipes

MartinPetkov:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/297


templates/tfengine/recipes/org/folder.hcl
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+data = {
+	parent = "organizations/{{.org_id}}"
```
umairidris, line 16: Doesn't look like this parent field is needed by terragrunt. Was this just for your testing?

MartinPetkov: That's correct. I looked for a place where I can pull it out and have it be used by both recipes, but I couldn't find one, so I just pulled it out in an arbitrary recipe.

umairidris: It is generally better to put data in a template if possible as it is clearer. The main appeal of top level data block is to avoid having to duplicate across many resources. If there is no example of this today I would prefer not manufacturing one just to test this behaviour. As we develop more templates or clean up our recipes we can use this where appropriate.

MartinPetkov: I'll remove it for now then. But it did work when I tested it.

xingao267:
LGTM

leave it to umair to approve

internal/tfengine/tfengine.go
```diff
@@ -86,7 +86,16 @@ func dump(conf *Config, root, outputPath string) error {
...
+			if rc.Data == nil {
+				rc.Data = make(map[string]interface{})
+			}
+			if err := template.MergeData(rc.Data, data, nil); err != nil {
+				return err
+			}
+
+			// Validate the schema, if present.
 			if len(rc.Schema) > 0 {
 				if err := jsonschema.ValidateMap(rc.Schema, data); err != nil {
```
umairidris, line 100: Does this need to be rc.Data?

MartinPetkov: I think so, good catch. Fixed.

PR #307, opened 2020-06-16
Flatten recipes, add more schemas

### LINKED ISSUE #304: Template overwrites not working properly
In engine templates, the template data does not correctly overwrite the top level data block defined in the config.

umairidris:
Fix #304 

#74


examples/tfengine/full.hcl
```diff
@@ -33,7 +33,7 @@ data = {
 }
 
 template "devops" {
-  recipe_path = "{{$base}}/org/devops.hcl"
+  recipe_path = "{{$base}}/devops.hcl"
```
MartinPetkov, line 36: Nit: Looking at it again, I think it would be clearer if the `$base` variable is named `$recipes`.

examples/tfengine/simple.hcl
```diff
@@ -28,7 +28,7 @@ data = {
 }
 
 template "devops" {
-  recipe_path = "{{$base}}/org/devops.hcl"
+  recipe_path = "{{$base}}/devops.hcl"
```
MartinPetkov, line 31: Nit: Same comment here, would prefer to rename the variable to `$recipes`.

templates/tfengine/recipes/folder.hcl
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+  {{end}}
+}
+
+template "folder" {
+  component_path = "../components/folder/folder"
+  output_path    = "{{.display_name}}/folder"
+  data = {
+    display_name = "{{.display_name}}"
```
MartinPetkov, line 67: Is a parent not needed when the parent is a folder?

umairidris: The top level data block sets parent_type to "organization"

examples/tfengine/full.hcl
umairidris: I like that. Changed

templates/tfengine/recipes/folder.hcl
umairidris: The parent will be set by the terragrunt template above

PR #312, opened 2020-06-17
Add resources schema

umairidris:
#74


templates/tfengine/recipes/resources.hcl
```diff
@@ -12,6 +12,584 @@
...
+            description = "Expiration in milliseconds."
+            type        = "integer"
+          }
+          access = {
+            description = <<EOF
+              Access for this bigquery dataset.
+              Each object should contain exactly one of the following keys:
+              - group_by_email: An email address of a Google Group to grant access to.
+              - user_by_email:  An email address of a user to grant access to.
+              - group_by_email: An email address of a Google Group to grant access to.
```
MartinPetkov, line 118: Duplicated line

umairidris: Good catch

PR #314, opened 2020-06-17
Add terraform addons schema

### LINKED ISSUE #74: Support schema for engine config and recipes

umairidris:
Fix #74


templates/tfengine/recipes/terragrunt_deployment.hcl
```diff
@@ -12,6 +12,85 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+schema = {
+  title       = "Terraform Deployment Recipe."
```
MartinPetkov, line 16: Shouldn't this be Terragrunt?

templates/tfengine/recipes/terragrunt_deployment.hcl
```diff
@@ -12,6 +12,85 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+schema = {
```
MartinPetkov, line 15: This schema looks duplicated. Is there no way to un-duplicate?

umairidris: I think in a follow up PR I will merge the two into one recipe called deployment and that will have a flag for enabling terragrunt.

templates/tfengine/recipes/terragrunt_deployment.hcl
umairidris: Right, fixed

PR #317, opened 2020-06-18
Simplify project recipe

umairidris:
Make project directory structure simpler and more consistent with org. Now the user can set the project directory to whatever they like (it doesn't need to be project ID like it was before). Continue using project ID in our examples.


umairidris:
> Do you need to regen the examples?

I ran the regen script to verify there is no diff from this change.

PR #323, opened 2020-06-19
Add golangci-lint to CI

MartinPetkov:
Inspired by https://github.com/github/super-linter. Super Linter itself produces too much output for our repo.

Also correct findings.


internal/tfengine/tfengine.go
```diff
@@ -141,6 +141,11 @@ func ConvertToLocalBackend(path string) error {
 
 	// Replace all GCS backend blocks with local.
 	fn := func(path string, info os.FileInfo, err error) error {
+		// Don't ignore the incoming error.
```
umairidris, line 144: I don't think comment adds much.

MartinPetkov: Removed.

PR #334, opened 2020-06-23
replace log export module with resource

umairidris:


MartinPetkov:
What's the issue with the module?

umairidris:
> What's the issue with the module?

It adds no benefit to using the resource. The only benefit it has is to support org/folder from thes ame module but our templates can produce the exact resource anyway. The submodules for individual destinations are not sufficient for our use and migrating them will take time.

So in the future we can reconsider this module, but for now given we aren't using the destination submodules (the main benefit), I don't see the benefit of using the top level module either.

MartinPetkov:
> > What's the issue with the module?
> 
> It adds no benefit to using the resource. The only benefit it has is to support org/folder from thes ame module but our templates can produce the exact resource anyway. The submodules for individual destinations are not sufficient for our use and migrating them will take time. And in order to support the logic for supporting multiple levels the logic becomes far more complicated within the module.
> 
> So in the future we can reconsider this module, but for now given we aren't using the destination submodules (the main benefit), I don't see the benefit of using the top level module either.

Thank, that makes sense. I just wanted the justification written down for posterity.

umairidris:
Definitely a good idea, in my rush I sometimes forget these things :)

PR #336, opened 2020-06-23
policygen: support reading multiple state file from dir

### LINKED ISSUE #318: Support reading state from a directory recursively

xingao267:
Fixes #318

### DIFF
--- a/cmd/policygen/main.go
+++ b/cmd/policygen/main.go
@@ -28,7 +28,7 @@ import (
 
 var (
 	configPath = flag.String("config_path", "", "Path to the Policy Generator config.")
-	statePath  = flag.String("state_path", "", "Path to Terraform state in json format.")
+	statePath  = flag.String("state_path", "", "Path to Terraform states in Terraform native .tfstate format. If a directory path is given, then all .tfstate files will be read recursively.")
 	outputPath = flag.String("output_path", "", "Path to output directory to write generated policies")
 )
 
--- a/internal/policygen/policygen_test.go
+++ b/internal/policygen/policygen_test.go
@@ -61,6 +61,20 @@ func TestExamples(t *testing.T) {
 		wantDirs  []wantDir
 	}{
 		{
+			// A single state file that does not have the .tfstate extension should still work.
+			"testdata/state",
+			[]wantDir{
+				{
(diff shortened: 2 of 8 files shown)


internal/policygen/policygen_test.go
```diff
@@ -49,90 +49,126 @@ func contains(s string, subs ...string) bool {
...
-					filepath.Join(tmp, "forseti_policies", "overall"),
+	tests := []struct {
+		configPath string
+		statePath  string
+		wantDirs   []wantDir
+	}{
+		{
+			configPath,
+			"testdata/org.tfstate",
+			append(
```
MartinPetkov, line 85: Since you're appending `common` for every test case, you should instead do that in the test loop below, not in the struct initialization.

internal/policygen/policygen_test.go
```diff
@@ -49,90 +49,126 @@ func contains(s string, subs ...string) bool {
...
-			}{
-				{
-					filepath.Join(tmp, "forseti_policies", "overall"),
+	tests := []struct {
+		configPath string
+		statePath  string
+		wantDirs   []wantDir
+	}{
+		{
+			configPath,
```
MartinPetkov, line 83: Same with `configPath`. This doesn't need to be part of the struct, since it's the same for all test cases.

xingao267: Done

internal/policygen/policygen_test.go
xingao267: Done

cmd/policygen/main.go
```diff
@@ -28,7 +28,7 @@ import (
 
 var (
 	configPath = flag.String("config_path", "", "Path to the Policy Generator config.")
-	statePath  = flag.String("state_path", "", "Path to Terraform state in json format.")
+	statePath  = flag.String("state_path", "", "Path to Terraform states in Terraform native .tfstate format. If a directory path is given, then all .tfstate files will be read recursively.")
```
umairidris, line 31: Does it read recursively in GCS buckets? That was actually my main FR

xingao267: Not yet. It's reading from a local dir first.

PR #339, opened 2020-06-23
make folder output path consistent with project

umairidris:
We stopped adding a prefix for projects, so do the same for folders.

### DIFF
--- a/templates/tfengine/recipes/folder.hcl
+++ b/templates/tfengine/recipes/folder.hcl
@@ -41,7 +41,7 @@ schema = {
 
 template "deployment" {
   recipe_path = "./deployment.hcl"
-  output_path = "{{.display_name}}/folder"
+  output_path = "./folder"
   data = {
     enable_terragrunt = true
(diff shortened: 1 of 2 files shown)


templates/tfengine/recipes/folder.hcl
```diff
@@ -41,7 +41,7 @@ schema = {
 
 template "deployment" {
   recipe_path = "./deployment.hcl"
-  output_path = "{{.display_name}}/folder"
+  output_path = "./folder"
```
xingao267, line 44: does these change the generated files?

umairidris: No it should generate the same files because i updated the output_path to include this prefix manually

xingao267: so in this case the {{display_name}} was .?

umairidris: It was "prod"

If you look at the example, template "folder_prod" now has the output path `./live/prod`. Before it was `./live`.

PR #343, opened 2020-06-24
Use higher-cpu machine to hopefully speed up CI

MartinPetkov:
Reference: https://cloud.google.com/cloud-build/docs/api/reference/rest/v1/projects.builds#machinetype


MartinPetkov:
Down to ~2.5min from ~8. Not going to use `N1_HIGHCPU_32` since it can make the job take noticeably longer to start.

PR #347, opened 2020-06-24
improve forseti cscc doc

xingao267:


umairidris:
You can also comment out the cscc field in the example and put a TODO on how to enable it similar to other commented out fields.

PR #352, opened 2020-06-24
Generalize the importer check script

MartinPetkov:


scripts/check_importer_supports_engine/main.go
```diff
@@ -33,7 +33,7 @@ import (
 	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/tfimport"
 )
 
-const full = "examples/tfengine/full.hcl"
+const examples = "examples/tfengine"
```
umairidris, line 36: examplesDir

scripts/check_importer_supports_engine/main.go
```diff
@@ -42,20 +42,45 @@ func main() {
 }
 
 func run() error {
-	if _, err := os.Stat(full); os.IsNotExist(err) {
-		return fmt.Errorf("example file %v does not exist: %v", full, err)
+	examples, err := filepath.Glob(filepath.Join(examples, "*.hcl"))
+	if err != nil {
+		return fmt.Errorf("filepath.Glob = %v", err)
```
umairidris, line 47: Doesn't matter too much given its an internal use script but in general you want to describe the func call for user facing errors instead of putting the func name (unlike tests where the func name helps). So this would be something like glob hcl files: %v

MartinPetkov: Done.

scripts/check_importer_supports_engine/main.go
MartinPetkov: Done.

PR #355, opened 2020-06-25
policygen: support reading state files from gcs

### LINKED ISSUE #206: Support reading state files from a GCS bucket.

xingao267:
Fixes #206 

TESTED=using our testing devops project

### DIFF
--- a/internal/policygen/load.go
+++ b/internal/policygen/load.go
@@ -103,8 +103,8 @@ func loadConfig(path string) (*config, error) {
 // loadResources loads Terraform state resources from the given path.
 // - If the path is a single local file, it loads resouces from it.
 // - If the path is a local directory, it walks the directory recursively and loads resources from each .tfstate file.
-// - If the path is a GCS bucket (indicated by 'gs://' prefix), it walks the bucket recursively and loads resources from each .tfstate file.
-//   It currently ignores the file/dir path inside the bucket and considers all .tfstate file in the bucket.
+// - If the path is a Cloud Storage bucket (indicated by 'gs://' prefix), it walks the bucket recursively and loads resources from each .tfstate file.
+//   It only reads the bucket name from the path and ignores the file/dir, if specified. All .tfstate file from the bucket will be read.
 func loadResources(path string) ([]*states.Resource, error) {
 	if strings.HasPrefix(path, "gs://") {
 		return loadResourcesFromCloudStorageBucket(path)
--- a/internal/policygen/load.go
+++ b/internal/policygen/load.go
@@ -15,18 +15,24 @@
 package policygen
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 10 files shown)


internal/policygen/load.go
```diff
@@ -94,33 +100,35 @@ func loadConfig(path string) (*config, error) {
...
 }
 
-// loadResources loads Terraform state resources from the given path. If the path is a single
-// file, it loads resouces from it. If the path is a directory, it walks the directory
-// recursively and loads resources from each .tfstate file.
+// loadResources loads Terraform state resources from the given path.
+// - If the path is a single local file, it loads resouces from it.
+// - If the path is a local directory, it walks the directory recursively and loads resources from each .tfstate file.
+// - If the path is a GCS bucket (indicated by 'gs://' prefix), it walks the bucket recursively and loads resources from each .tfstate file.
+//   It currently ignores the file/dir path inside the bucket and considers all .tfstate file in the bucket.
```
MartinPetkov, line 107: Is it difficult to consider the file or dir in the bucket?

xingao267: It's not difficult but just will make the code a bit messy and we currently don't need that. We can add that when there is a FR for that.

internal/policygen/load.go
```diff
@@ -91,36 +97,39 @@ func loadConfig(path string) (*config, error) {
...
 	if !fi.IsDir() {
-		resources, err := terraform.ResourcesFromState(path)
-		if err != nil {
-			return nil, fmt.Errorf("read resources from Terraform state file %q: %v", path, err)
-		}
-		return resources, nil
+		return loadResourcesFromSingleLocalFile(path)
 	}
 
 	var allResources []*states.Resource
```
umairidris, line 124: Consider moving this to a function e.g. loadResourcesFromDir.

internal/policygen/load.go
MartinPetkov: sg

internal/policygen/load.go
```diff
@@ -91,36 +97,39 @@ func loadConfig(path string) (*config, error) {
...
 	fn := func(path string, _ os.FileInfo, _ error) error {
```
umairidris, line 125: Check error?

internal/policygen/load.go
```diff
@@ -130,5 +139,66 @@ func loadResources(path string) ([]*states.Resource, error) {
...
+	}
+	return resources, nil
+}
+
+func loadResourcesFromCloudStorageBucket(path string) ([]*states.Resource, error) {
+	// Trim the 'gs://' prefix and split the path into the bucket name and cloud storage file path.
+	bucketName := strings.SplitN(strings.TrimPrefix(path, "gs://"), "/", 2)[0]
+	log.Printf("Reading state files from Cloud Storage bucket %q", bucketName)
+
+	ctx := context.Background()
```
umairidris, line 159: Pass context from top level caller

internal/policygen/load.go
xingao267: Done

PR #359, opened 2020-06-25
devops: support terraform only mode

### LINKED ISSUE #358: Support a Terraform-only mode in devops recipe.

xingao267:
Fixes #358


templates/tfengine/recipes/devops.hcl
```diff
@@ -69,10 +76,17 @@ template "bootstrap" {
...
 }
 
+{{if get . "enable_terragrunt" true}}
 template "root" {
   component_path = "../components/terragrunt/root"
   output_path    = "./live"
 }
+{{else}}
+template "root" {
+  component_path = "../components/terraform/root"
```
umairidris, line 86: Do we need a terraform root? I think in this case we should only generate the bootstrap and cicd dirs and leave the rest.

templates/tfengine/recipes/devops.hcl
```diff
@@ -83,6 +97,13 @@ template "cicd_manual" {
   }
 }
 
+{{if get . "enable_terragrunt" true}}
```
umairidris, line 100: Just something to think about but instead of enabling or disabling terragrunt we can introduce a variable "terragrunt_root"

In our templates we can set it to `terragrunt_root = "live"`. If that field is missing we don't generate Terragrunt root. It may make it more obvious where the 'live' value is coming from. WDYT?

templates/tfengine/recipes/devops.hcl
xingao267: yea, we actually still need it. The cloud build config need that to correctly determines where to find the terraform configs and its own config.

templates/tfengine/recipes/devops.hcl
xingao267: Yea that's a good idea. I like making the value "live" more explicit and obvious. I think it's more like a cicd_root, and terragrunt or terraform can both act on the configs in the root. We can discuss it more in our sync.

templates/tfengine/recipes/devops.hcl
xingao267: oh sorry... I misunderstood. You mean the terraform root component here. I think having some terraform only root files can still be helpful. I added this terraform root to include a README file for terraform only useage and a main.tf which sets the backend and people can start adding their configs in that file.

templates/tfengine/recipes/devops.hcl
umairidris: I still see a cicd in the live dir. Is that supposed to be there?

https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/tree/devops/examples/tfengine/generated/devops/live/cicd

templates/tfengine/recipes/devops.hcl
umairidris: Does the solutions doc talk about how to manage multple deployments? Or are you supposed to put everything under the single live deployment?

Another caveat is all of our other templates will still continue to generate terragrunt.hcl files which will stop working, so we may need to think about this change a bit more. Worth a short meeting to discuss?

xingao267: Sure let’s chat tomorrow.

templates/tfengine/recipes/devops.hcl
xingao267: Resolved offline. Added a deployment folder.

templates/tfengine/recipes/devops.hcl
xingao267: Resolved offline.

xingao267:
TESTED=using a new devops project in our testing org and deploy some native TF resources using this pipeline.

Issue #356, opened 2020-06-25
Format generated files


umairidris:

xingao267:
#364

xingao267:
Is this one done?

umairidris:
I still need to port over terraform-clean-syntax so that we convert "${foo}" to just foo.

xingao267:
got it

umairidris:
Re-opening as we still need to clean up empty lines

xingao267:
@umairidris can we also clean up empty lines before 0.2.0 release?

umairidris:
Sure I'll spend some time today looking into this.

PR #360, opened 2020-06-25
format generated engine files

umairidris:
#356

### DIFF
--- a/internal/hcl/hcl.go
+++ b/internal/hcl/hcl.go
@@ -17,11 +17,33 @@ package hcl
 import (
 	"encoding/json"
 	"fmt"
+	"os/exec"
 
+	"github.com/GoogleCloudPlatform/healthcare/deploy/runner"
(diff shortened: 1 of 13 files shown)


internal/tfengine/tfengine.go
```diff
@@ -168,3 +174,21 @@ func ConvertToLocalBackend(path string) error {
 
 	return nil
 }
+
+func formatDir(dir string) error {
```
xingao267, line 178: move this to a common location? I can use it to format org policies as well in policygen

internal/hcl/hcl.go
```diff
@@ -17,12 +17,34 @@ package hcl
 import (
 	"encoding/json"
 	"fmt"
+	"os/exec"
 
+	"github.com/GoogleCloudPlatform/healthcare/deploy/runner"
```
xingao267, line 22: It's using our older runner lolll

umairidris: sneaky

Issue #328, opened 2020-06-22
Feature parity with DPT


gbhojak:
Hi,

I would like to understand how does this new suite compare with existing DPT solution [here](https://github.com/GoogleCloudPlatform/healthcare).

I have been reading [this guide ](https://cloud.google.com/solutions/architecture-hipaa-aligned-project) and [this one as well](https://cloud.google.com/solutions/setting-up-a-hipaa-aligned-project) to setup our infrastructure for HIPPA.
We are trying to build an infrastructure that closely resembles [this](https://cloud.google.com/solutions/architecture-hipaa-aligned-project#architecture_diagram).

Looking at the full.hcl example, it does look like that with this approach each of our environments such as staging and production would have their own devops/audit/monitoring GCP projects. Is this the right assumption?
Also, would it make to have multiple copies of full.hcl and create one for each env, like full-staging and full-prod or something like that?

I am really excited to try this out for our needs. Just need to understand above differences.
Apologies for asking so many questions in one go. Just getting started with this suite and in the process of creating staging and production envs.

Regards,
Giriraj

umairidris:
Hi Giriraj,

Thanks for your interest! In terms of features, I would say the new toolkit is relatively on par with DPT. It in fact supports more features such as org level auditing/monitoring and networking resources natively. Where things are lacking are documentations and clearer examples which we are actively working on. 

We will soon be adding some features to support folder level management for things like devops, etc (https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/40).

With the above we will be updating the examples and documentation to better architect organizations.

In general we will recommend:
- one "org" config per org for managing org level auditing and monitoring
- one "folder" config per team/environment

Each of the above will have its own devops project and CICD pipeline.

The folder boundary will depend on each org. For smaller orgs, having a "prod" folder (and thus prod devops) will be fine. So they would have the config files like org_foundation.hcl, folder_prod.hcl, folder_dev.hcl. If it should be for each team/app then you may have something like org_foundation.hcl, folder_prod_team1.hcl, folder_dev_team1.hcl, etc...

We will update the README and examples to reflect the above changes.

I will be sure to include relevant PRs to this issue so you can track them and give feedback.

umairidris:
One thing to keep in mind is you may experience more breaking changes in the next couple of months as we figure out the best way to deploy these complex infra. Thus whenever you sync to head please reference the latest examples to make sure you are not deviating from them.

[...]

PR #363, opened 2020-06-25
create some sub folders in folder_foundation

umairidris:
#328

### DIFF
--- a/examples/tfengine/generated/folder_foundation/live/prod/team1/folder/terragrunt.hcl
+++ b/examples/tfengine/generated/folder_foundation/live/prod/team1/folder/terragrunt.hcl
@@ -1,0 +1,30 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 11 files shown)


examples/tfengine/generated/folder_foundation/live/prod/team1/folder/terragrunt.hcl
```diff
@@ -0,0 +1,30 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include {
+  path = find_in_parent_folders()
+}
+
+dependency "parent_folder" {
+  config_path = "../../folder"
```
xingao267, line 20: I wonder is this config_path always holds? Can users put the parent folder in a wrong directory and this is not valid?

umairidris: Yeah that's a good question.. in that case I guess it just wouldn't work and the user may need to set their own custom terraform_addon type thing. We can support that if the need arises, but this way kind of forces our layout standardization which can be useful.

PR #364, opened 2020-06-26
policygen: format generated tf files

xingao267:
also make fmt commands and output quiet

### DIFF
--- a/internal/hcl/hcl.go
+++ b/internal/hcl/hcl.go
@@ -27,18 +27,18 @@ import (
 
 // FormatDir formats the hcl files in the given directory.
 // The user must have terraform and terragrunt binaries in their PATH.
-func FormatDir(dir string) error {
-	rn := &runner.Default{}
-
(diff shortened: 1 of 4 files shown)


internal/hcl/hcl.go
```diff
@@ -27,18 +27,18 @@ import (
...
 // FormatDir formats the hcl files in the given directory.
 // The user must have terraform and terragrunt binaries in their PATH.
-func FormatDir(dir string) error {
-	rn := &runner.Default{}
-
+func FormatDir(rn runner.Runner, dir string) error {
 	tfFmt := exec.Command("terraform", "fmt", "-recursive")
 	tfFmt.Dir = dir
-	if err := rn.CmdRun(tfFmt); err != nil {
+	// Silent stdout.
```
umairidris, line 33: Silence

xingao267: Done

PR #368, opened 2020-06-29
improve cloud build trigger attributes

xingao267:


templates/tfengine/recipes/devops.hcl
```diff
@@ -143,6 +122,57 @@ schema = {
...
         }
+        validate_trigger = {
+          description = <<EOF
+            Config block for the presubmit validation Cloud Build trigger. If specified, create
+            the trigger and grant the Cloud Build Service Account necessary permissions to perform
+            the build.
+          EOF
+          type        = "object"
+          properties = {
+            disable = {
```
umairidris, line 133: On a closer look I think it may be worth renaming this to enable actually. Reason is I want to rename bootstrap_gcs_backend to enable_bootstrap_gcs_backend and having a consistent naming scheme for all templates might be useful.

Optional though so if you prefer this I am fine with it too

xingao267: Discussed offline. Will leave it as disable

Issue #295, opened 2020-06-15
Engine schema should be more restrictive and do not allow additional properties


xingao267:
Engine should also set `additionalProperties = false` in schemas and make the validation more restrictive. This gives users more directions and error messages when their schema structure is wrong, or has typos or bad keywords. Otherwise, the engine might just run successfully but their engine configs can be written wrong, and changes didn't get picked up by the engine.

xingao267:
I think this should have higher priority and start being looked into. This can help users a lot in understanding what went wrong and reduce silent errors.

umairidris:
This is a bit non trivial because if we were to diasllow extra fields suddenly then there will be A LOT of duplication. Imagine every project having to duplicate the parent_id or location being set on every resource. We need a way to still have a list of 'common' top level data block that is acceptable

PR #376, opened 2020-07-02
Disallow additional properties

umairidris:
This only checks the template data (not the merged data)
This currently does not prevent additional properties for fields that use `flatten`. I will fix that in a future CL

#295


internal/tfengine/tfengine.go
```diff
@@ -71,64 +71,75 @@ func Run(confPath, outPath string, format bool) error {
...
-			cp, err := pathutil.Expand(ti.ComponentPath)
-			if err != nil {
+		if err := dump(rc, filepath.Dir(rp), outputPath); err != nil {
+			return fmt.Errorf("recipe %q: %v", rp, err)
+		}
+
+	case ti.ComponentPath != "":
+		if ti.Name == "org_policies" {
+			// Only check against unmerged template data so we can disallow additional properties in the schema.
+			if err := policygen.ValidateOrgPoliciesConfig(ti.Data, true); err != nil {
```
xingao267, line 130: You can remove the second field so it always disallow additional properties https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/internal/policygen/load.go#L48

umairidris: Done.

umairidris:
I will be making it tighter in a follow up, since we use flattening in a lot of recipes and that isn't being checked!

PR #380, opened 2020-07-03
Make base templates more scalable through use of objects

umairidris:
Note: state_bucket is a bit of an exception where we won't create an object because it is used throughout many different templates and thus converting it to an object actually makes the templates more confusing.


examples/tfengine/devops.hcl
```diff
@@ -27,23 +28,24 @@ template "devops" {
...
-      "group:example-devops-owners@example.com",
-    ]
-
-    enable_terragrunt = false,
+    admins_group      = "example-org-admin@example.com"
+    state_bucket      = "example-terraform-state"
+    enable_terragrunt = false
 
+    project = {
+      project_id = "example-devops"
```
MartinPetkov, line 36: How come the "project" prefix stays for id, but not for owners?

umairidris: I stuck with using project_id (and not 'id') as that is what the underlying terraform resource uses: https://www.terraform.io/docs/providers/google/r/google_project.html.

Same with BQ dataset. The resource calls the main id 'dataset_id'. So I prefer to be consistent with the API.

examples/tfengine/devops.hcl
```diff
@@ -27,23 +28,24 @@ template "devops" {
...
-    project_id       = "example-devops"
-    state_bucket     = "example-terraform-state"
-    storage_location = "us-central1"
-    admins_group     = "example-org-admin@example.com"
-    project_owners = [
-      "group:example-devops-owners@example.com",
-    ]
-
-    enable_terragrunt = false,
+    admins_group      = "example-org-admin@example.com"
```
xingao267, line 31: admins

templates/tfengine/recipes/monitor.hcl
```diff
@@ -16,14 +16,42 @@ schema = {
...
+          description = "ID of project."
+          type        = "string"
+        }
+      }
     }
-    domain = {
-      description = "Domain for the Forseti instance."
-      type        = "string"
+    forseti = {
+      description = "Config for the Forseti instance."
```
xingao267, line 31: additionalProperties = false

examples/tfengine/devops.hcl
umairidris: Done

templates/tfengine/recipes/monitor.hcl
umairidris: Done

PR #382, opened 2020-07-03
fix master_authorized_networks block not picked up by engine

xingao267:


templates/tfengine/components/resources/gke_clusters/main.tf
```diff
@@ -29,8 +29,17 @@ module "{{resourceName . "name"}}" {
   ip_range_services      = "{{.ip_range_services_name}}"
   master_ipv4_cidr_block = "{{.master_ipv4_cidr_block}}"
 
-  # Optional.
-  master_authorized_networks = var.master_authorized_networks
+  {{- if has . "master_authorized_networks"}}
+  master_authorized_networks = [
+    {{- range .master_authorized_networks}}
```
MartinPetkov, line 34: Can you use any of the hcl/hclField functions here? Seems like a list of objects should expand correctly.

MartinPetkov: I just tried it. You can do
```
master_authorized_networks = {{hcl .master_authorized_networks}}
```

And the generated files are the same.

umairidris: +1 I think you should be able to do {{hclField . "master_authorized_networks" false}}

xingao267: Done!

PR #386, opened 2020-07-06
Support checking additional properties for flattened templates

### LINKED ISSUE #295: Engine schema should be more restrictive and do not allow additional properties

umairidris:
Fix #295


internal/policygen/schema.go
```diff
@@ -64,11 +64,6 @@ properties = {
 var orgPoliciesSchema = []byte(`
 title                = "Google Cloud Organization Policies Config Schema"
 additionalProperties = false
-required = [
-  "parent_type",
-  "parent_id",
-  "allowed_policy_member_customer_ids",
```
xingao267, line 70: do we still want `allowed_policy_member_customer_ids` to be required?

templates/tfengine/recipes/resources.hcl
```diff
@@ -104,6 +104,13 @@ schema = {
             description = "ID of bigquery dataset."
             type        = "string"
           }
+          resource_name = {
+            description = <<EOF
+              Override for Terraform resource name. If unset, defaults to normalized dataset_id.
+              Normalization will occur based on the
```
xingao267, line 110: based on the ...

internal/policygen/schema.go
umairidris: Done

templates/tfengine/recipes/resources.hcl
umairidris: Fixed

PR #387, opened 2020-07-07
support folder audit config

### LINKED ISSUE #327: Support google_folder_iam_audit_config in engine and google provider

xingao267:
Fixes #327

### DIFF
--- a/templates/tfengine/components/audit/main.tf
+++ b/templates/tfengine/components/audit/main.tf
@@ -23,10 +23,10 @@ terraform {
 }
 
 {{- $parent_field := "org_id"}}
-{{- $parent_var := printf "\"%s\"" .parent_id}}
+{{- $parent_var := "var.org_id"}}
 {{- if eq .parent_type "folder"}}
   {{- $parent_field = "folder"}}
-  {{- $parent_var = printf "\"folders/%s\"" .parent_id}}
+  {{- $parent_var = "var.folder"}}
 {{- end}}
 
 # IAM Audit log configs to enable collection of all possible audit logs.
--- a/templates/tfengine/components/audit/main.tf
+++ b/templates/tfengine/components/audit/main.tf
@@ -23,14 +23,12 @@ terraform {
 }
 
 {{- $parent_field := "org_id"}}
-{{- $parent_var := "var.org_id"}}
+{{- $parent_var := printf "\"%s\"" .parent_id}}
 {{- if eq .parent_type "folder"}}
 {{- $parent_field = "folder"}}
-{{- $parent_var = "var.folder_id"}}
+{{- $parent_var = printf "\"folders/%s\"" .parent_id}}
 {{- end}}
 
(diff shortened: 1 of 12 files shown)


templates/tfengine/components/audit/main.tf
```diff
@@ -23,14 +23,12 @@ terraform {
 }
 
 {{- $parent_field := "org_id"}}
-{{- $parent_var := "var.org_id"}}
```
umairidris, line 26: The reason I preferred to use vars here was to reduce duplication a bit in the generated configs. Should we keep the vars?

templates/tfengine/components/audit/main.tf
```diff
@@ -23,14 +23,12 @@ terraform {
...
+{{- $parent_var := printf "\"%s\"" .parent_id}}
 {{- if eq .parent_type "folder"}}
 {{- $parent_field = "folder"}}
```
umairidris, line 28: Can you indent these fields a bit so it's clear it's within the if

templates/tfengine/components/audit/main.tf
```diff
@@ -23,14 +23,12 @@ terraform {
...
-{{- $parent_var = "var.folder_id"}}
+{{- $parent_var = printf "\"folders/%s\"" .parent_id}}
```
umairidris, line 29: Before we were using the folder id directly and now it's folders/. Is that OK for all resources within this module?

templates/tfengine/components/audit/main.tf
xingao267: Done

templates/tfengine/components/audit/main.tf
xingao267: yes

https://www.terraform.io/docs/providers/google/r/google_folder_iam_member.html
https://www.terraform.io/docs/providers/google/r/logging_folder_sink.html

templates/tfengine/components/audit/main.tf
xingao267: Could you elaborate a little on how it reduce duplication in the generated configs?

And also in a lot of places we are directly use values in `main.tf`, not through the `terraform.tfvars`.

umairidris: Yeah in general I think we are now preferring to hard code values within main.tf. I have still preferred to use variables for the primary parent id, for example all resources still use var.project_id within projects: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/examples/tfengine/generated/team/live/example-prod-data/resources/main.tf#L32.

Since parent ids are static throughout the deployment it's relatively safe to make them vars so that you don't see the number repeated so many times throughout the deployment. However if you want to just have one way of doing things and just putting the id directly in main.tf that's fine with me too.

xingao267: Discussed offline. Use vars.

PR #390, opened 2020-07-07
fix cloud sql schema and extra fields

xingao267:


examples/tfengine/team.hcl
```diff
@@ -239,7 +239,6 @@ template "project_data" {
           type               = "mysql"
           network_project_id = "example-prod-networks"
           network            = "example-network"
-          subnet             = "example-subnet"
```
MartinPetkov, line 242: Is subnet not used for Cloud SQL instances?

templates/tfengine/recipes/resources.hcl
```diff
@@ -166,6 +168,11 @@ schema = {
             description = "Name of the cloud sql instance."
             type        = "string"
           }
+          type = {
```
umairidris, line 171: nit: Move type below resource_name. I would like to group resource_name with the field it uses by default if not present (e.g. name).

examples/tfengine/team.hcl
xingao267: it's not

templates/tfengine/recipes/resources.hcl
xingao267: Done

PR #391, opened 2020-07-07
fix subnet private access typo

xingao267:
https://github.com/terraform-google-modules/terraform-google-network

### DIFF
--- a/examples/tfengine/generated/team/live/example-prod-networks/resources/main.tf
+++ b/examples/tfengine/generated/team/live/example-prod-networks/resources/main.tf
@@ -58,18 +58,18 @@ module "example_network" {
   project_id   = var.project_id
   subnets = [
     {
-      subnet_name            = "example-bastion-subnet"
-      subnet_ip              = "10.1.0.0/16"
-      subnet_region          = "us-central1"
-      subnet_flow_logs       = true
-      subnets_private_access = true
+      subnet_name           = "example-bastion-subnet"
+      subnet_ip             = "10.1.0.0/16"
+      subnet_region         = "us-central1"
+      subnet_flow_logs      = true
+      subnet_private_access = true
     },
     {
-      subnet_name            = "example-gke-subnet"
-      subnet_ip              = "10.2.0.0/16"
-      subnet_region          = "us-central1"
-      subnet_flow_logs       = true
-      subnets_private_access = true
+      subnet_name           = "example-gke-subnet"
+      subnet_ip             = "10.2.0.0/16"
+      subnet_region         = "us-central1"
+      subnet_flow_logs      = true
+      subnet_private_access = true
     },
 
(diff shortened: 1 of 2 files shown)


templates/tfengine/components/resources/compute_networks/main.tf
```diff
@@ -31,7 +31,7 @@ module "{{$resource_name}}" {
       subnet_ip              = "{{.ip_range}}"
       subnet_region          = "{{get . "compute_region" $.compute_region}}"
       subnet_flow_logs       = true
-      subnets_private_access = true
+      subnet_private_access = true
```
umairidris, line 34: Isn't this supposed to be {{get . subnet_private_access false}} ?

examples/tfengine/generated/team/live/example-prod-networks/resources/main.tf
```diff
@@ -58,18 +58,18 @@ module "example_network" {
   project_id   = var.project_id
   subnets = [
     {
-      subnet_name            = "example-bastion-subnet"
-      subnet_ip              = "10.1.0.0/16"
-      subnet_region          = "us-central1"
-      subnet_flow_logs       = true
-      subnets_private_access = true
```
umairidris, line 65: Did the schema check miss this?

templates/tfengine/components/resources/compute_networks/main.tf
```diff
@@ -31,7 +31,7 @@ module "{{$resource_name}}" {
       subnet_ip              = "{{.ip_range}}"
       subnet_region          = "{{get . "compute_region" $.compute_region}}"
       subnet_flow_logs       = true
```
umairidris, line 33: Hmm.. even subnet_flow_logs is always true so the value in the config seems to be ignored

examples/tfengine/generated/team/live/example-prod-networks/resources/main.tf
```diff
@@ -58,18 +58,18 @@ module "example_network" {
...
+      subnet_name           = "example-bastion-subnet"
+      subnet_ip             = "10.1.0.0/16"
+      subnet_region         = "us-central1"
+      subnet_flow_logs      = true
+      subnet_private_access = true
```
umairidris, line 65: Do you know why this missed our test? Shouldn't terraform plan fail?

umairidris: Nevermind, subnets are implemented as a map:https://github.com/terraform-google-modules/terraform-google-network/blob/master/variables.tf#L38

xingao267: Yea, I think once the named object is there, we can make it more restrict as well in the CFT module.

PR #393, opened 2020-07-07
Make hclField always optional

### LINKED ISSUE #384: Make hclField always optional
We should only use hclField for optional fields. Thus, we can remove the boolean 'required' key in the func call.

Before:
`{{hclField . "foo" false}}`

After:
`{{hclField . "foo"}}`

For required fields use hcl:
`foo = {{hcl .foo}}`

umairidris:
Fix #384

### DIFF
--- a/internal/template/funcmap.go
+++ b/internal/template/funcmap.go
@@ -72,21 +72,17 @@ func hcl(v interface{}) (string, error) {
 	return string(b), nil
 }
 
-// hclField returns a hcl marshaled field e.g. `name = "foo"`
-func hclField(m map[string]interface{}, key string, req bool) (string, error) {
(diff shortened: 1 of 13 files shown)


internal/template/funcmap.go
```diff
@@ -72,21 +72,17 @@ func hcl(v interface{}) (string, error) {
 	return string(b), nil
 }
 
-// hclField returns a hcl marshaled field e.g. `name = "foo"`
-func hclField(m map[string]interface{}, key string, req bool) (string, error) {
+// hclField returns a hcl marshaled field e.g. `name = "foo"`, if present.
```
xingao267, line 75: maybe add a note if want to enforce the field to be present, use `hcl`

umairidris: Good idea, done

PR #394, opened 2020-07-07
Run validate-all in engine tests

MartinPetkov:
This will also run init, like plan-all, but will additionally validate that the generated configs.


umairidris:
I am not sure what this is adding. Doesn't plan-all cover validation?

MartinPetkov:
I didn't realize plan already runs validate. I just checked [here](https://www.terraform.io/docs/commands/validate.html) and saw "use the terraform plan command instead, which includes an implied validation check". I'll close.

Issue #397, opened 2020-07-08
Remove need for Terragrunt


umairidris:

umairidris:
This is being tracked in the no_terragrunt branch.

#399 #401 #405 #429 #431 #454 #459

PR #399, opened 2020-07-08
update tfengine test to only use terraform

umairidris:
#397


internal/tfengine/tfengine_test.go
```diff
@@ -48,18 +48,40 @@ func TestExamples(t *testing.T) {
...
+			// Run plan to verify configs.
+			fs, err := ioutil.ReadDir(tmp)
+			if err != nil {
+				t.Fatalf("ioutil.ReadDir = %v", err)
+			}
 
-			// Convert the configs not reference a GCS backend as the state bucket does not exist.
-			if err := ConvertToLocalBackend(path); err != nil {
-				t.Fatalf("ConvertToLocalBackend(%v): %v", path, err)
+			// Skip these dirs as they use data sources that are not available until dependencies have been deployed.
```
MartinPetkov, line 57: Wouldn't this mean that we're no longer testing configs under these directories? What happens if we make a mistake and these generate invalid configs?

umairidris: We never tested them, only the 'live' dir. Since now we are flattening things live stuff will move to top level and get tested automatically.

xingao267:
Can we actually test both workflow? For now, we are still using terragrunt, and not sure if we should just drop the test with terragrunt. Maybe do both?

umairidris:
> Can we actually test both workflow? For now, we are still using terragrunt, and not sure if we should just drop the test with terragrunt. Maybe do both?

This is going into a new branch (no_terragrunt), not master. By the time we merge no_terragrunt into master there won't be any terragrunt code left in the codebase.

xingao267:
ohhhh didn't see the target branch.

PR #405, opened 2020-07-09
Flatten all directories into single top level ones

umairidris:
Note: the configs are still not functional, but this is what they will look like in the end.


umairidris:
CICD is curently the most broken because it relied so heavily on terragrunt. I think we will need a cohesive look at CICD later to make it work in this new paradigm.

umairidris:
@xingao267 just an FYI I just moved the cicd to its own template for now. The main reason is I want to explicitly state the output directory for each template is so that when we need to list it in the CICD it is obvious what directories we need to list. Let's talk about it in our sync.

PR #406, opened 2020-07-09
fix most references

umairidris:


examples/tfengine/generated/org_foundation/audit/outputs.tf
```diff
@@ -1,17 +1,3 @@
-# Copyright 2020 Google LLC
```
MartinPetkov, line 1: This lost the license header, we should put it back.

MartinPetkov: Same in some other files lower down.

umairidris: An unfortunate result of merging deployments together. I will use xin's helper to add the headers instead of trying to do it manually ourselves in our templates to fix this later.

PR #408, opened 2020-07-10
Add support for compute instance templates and instances

### LINKED ISSUE #204: Support compute instances and bastion hosts
```
module "bastion" {
  source = "terraform-google-modules/bastion-host/google"

  name         = "bastion-vm"
  host_project = var.project_id
  project      = var.project_id
  region       = var.region
  zone         = var.zone
  network      = "foo"
  subnet       = "foo"
  image_family = "ubuntu-1804-lts"
  members      = var.bastion_users
  scopes = [
    "https://www.googleapis.com/auth/sqlservice.admin",
  ]
  startup_script = <<EOF
#!/bin/bash
sudo apt-get -y update
sudo apt-get -y install mysql-client-core-5.7
sudo wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 -O /usr/local/bin/cloud_sql_proxy
sudo chmod +x /usr/local/bin/cloud_sql_proxy
EOF
}
```

umairidris:
Fix #204

### DIFF
--- a/examples/tfengine/generated/team/live/example-prod-apps/resources/main.tf
+++ b/examples/tfengine/generated/team/live/example-prod-apps/resources/main.tf
@@ -38,7 +38,6 @@ module "example_instance_template" {
     email  = "${google_service_account.example_sa.email}"
     scopes = ["cloud-platform"]
   }
-
   enable_shielded_vm = true
   shielded_instance_config = {
     enable_secure_boot          = true
--- a/examples/tfengine/generated/team/live/example-prod-apps/resources/main.tf
+++ b/examples/tfengine/generated/team/live/example-prod-apps/resources/main.tf
@@ -23,6 +23,39 @@ terraform {
 }
 
 
+
+module "example_instance_template" {
+  source  = "terraform-google-modules/vm/google//modules/instance_template"
+  version = "~> 3.0.0"
+
+  name_prefix        = "example-instance-template"
+  project_id         = var.project_id
+  region             = "us-central1"
+  subnetwork_project = "example-prod-networks"
+  subnetwork         = "example-instance-subnet"
+
+  service_account = {
+    email  = "${google_service_account.example_sa.email}"
+    scopes = ["cloud-platform"]
+  }
+
+  enable_shielded_vm = true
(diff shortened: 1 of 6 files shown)


templates/tfengine/components/resources/compute_instance_templates/main.tf
```diff
@@ -0,0 +1,67 @@
...
+
+  {{- hclField . "machine_type"}}
+  {{- hclField . "image_family"}}
+  {{- hclField . "image_project"}}
+  {{- hclField . "disk_size_gb"}}
+  {{- hclField . "disk_type"}}
+  {{- hclField . "preemptible"}}
+
+  service_account = {
+    email = "{{.service_account}}"
```
xingao267, line 37: is there a way to auto-allocated default SA?

examples/tfengine/generated/team/live/example-prod-apps/resources/main.tf
```diff
@@ -23,6 +23,39 @@ terraform {
...
+  region             = "us-central1"
+  subnetwork_project = "example-prod-networks"
+  subnetwork         = "example-instance-subnet"
+
+  service_account = {
+    email  = "${google_service_account.example_sa.email}"
+    scopes = ["cloud-platform"]
+  }
+
+  enable_shielded_vm = true
```
xingao267, line 42: This actually puts limitation on the image family. If they want to use a unsupported image family, this might fail.

templates/tfengine/components/resources/compute_instance_templates/main.tf
umairidris: It would become the default compute SA, so I think passing an SA should be a good best practice. It's also what's recommended here: https://cloud.google.com/compute/docs/access/service-accounts#associating_a_service_account_to_an_instance

examples/tfengine/generated/team/live/example-prod-apps/resources/main.tf
umairidris: Made it overridable.

PR #413, opened 2020-07-13
Add time function and function to add int

MartinPetkov:
Useful for generating the current time or expiry dates (via `(Time).AddDate`).

### DIFF
--- a/internal/template/funcmap.go
+++ b/internal/template/funcmap.go
@@ -17,6 +17,7 @@ package template
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/rodaine/hclencoder"
 )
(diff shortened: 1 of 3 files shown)


internal/template/funcmap.go
```diff
@@ -28,6 +29,8 @@ var funcMap = map[string]interface{}{
 	"hclField":     hclField,
 	"replace":      replace,
 	"resourceName": resourceName,
+	"now":          time.Now,
```
xingao267, line 32: what's the use case for these?

internal/template/funcmap.go
```diff
@@ -111,3 +114,12 @@ func resourceName(m map[string]interface{}, key string) (string, error) {
 func replace(s, old, new string) string {
 	return strings.Replace(s, old, new, -1)
 }
+
+// add adds a list of integers together
+func add(nums ...int) int {
```
umairidris, line 119: sum

internal/template/funcmap.go
MartinPetkov: It would be useful for generating expiry dates.

internal/template/funcmap.go
MartinPetkov: Removed.

Issue #416, opened 2020-07-13
Allow template files with .tmpl extension


MartinPetkov:
Some files may need to be generated with a different name, but can't be checked-in with that name. An example might be an OWNERS file, or .gitattributes. Even though we don't have that use case right now, it's useful to support it.

Simple solution: if any component file has the extension `.tmpl`, remove it before writing it out.

PR #417, opened 2020-07-13
Remove .tmpl ext before writing component file

MartinPetkov:
Fix https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/416

### DIFF
--- a/internal/template/template.go
+++ b/internal/template/template.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"text/template"
 
 	"github.com/imdario/mergo"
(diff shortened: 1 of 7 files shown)


internal/template/template.go
```diff
@@ -84,6 +88,12 @@ func WriteFile(in, out string, data map[string]interface{}) error {
 		return fmt.Errorf("parse template %q: %v", in, err)
 	}
 
+	// Rewrite if the file name has a template extension.
+	parts := strings.Split(out, ".")
```
umairidris, line 92: use filepath.Ext

MartinPetkov: Done. Followed the suggestion here: https://stackoverflow.com/a/13027975

umairidris: I think that makes more sense if you want to trim the ext always. In this case since we know the exact suffix I think the best approach is strings.TrimSuffix as that will work just as well and does not create an if or need for any indexing.

MartinPetkov: Oh I see what you mean. Done.

PR #419, opened 2020-07-13
upgrade dns module version

xingao267:


umairidris:
This should not be needed. ~> 3.0.0 should automatically fetch 3.0.1. Is it not working as intended?

xingao267:
It's actually fetching 3.0.2. Nvm.

PR #421, opened 2020-07-13
Basic build script

MartinPetkov:
Will use this for releasing the first version. Working on basic release docs too.

### DIFF
--- a/build/build-release.sh
+++ b/build/build-release.sh
@@ -17,7 +17,7 @@
 # Helper to build release artifacts - binaries, and templates and policies bundles.
 #
 # Usage (from repo root):
-# ./scripts/build.sh -o . -v v0.1.0
+# ./build/build-release.sh -o . -v v0.1.0
 
 set -e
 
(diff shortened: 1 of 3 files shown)


umairidris:
Should this be under build/ ?

Maybe build/release.sh ?

MartinPetkov:
> Should this be under build/ ?
> 
> Maybe build/release.sh ?

Good idea.

MartinPetkov:
> Should this be under build/ ?
> 
> Maybe build/release.sh ?

I wouldn't call it release.sh though. I think that's too vague and people may assume it's actually creating tags or GitHub releases too, when it's only building artifacts.

build/build-release.sh
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+version_regex="^v[0-9]+\.[0-9]+\.[0-9]+$"
+if ! [[ ${VERSION} =~ ${version_regex} ]]; then
+  echo "Version \"${VERSION}\" does not match regex ${version_regex}"
+  exit 1
+fi
+
+# Build binaries
+SUPPORTED_OS=("linux" "darwin" "windows")
```
xingao267, line 49: Just linux and darwin? since we didn't verify windows

MartinPetkov: I just tested it on Windows (using `scripts/regen.sh` but converted to run in Windows `cmd`). Seems to work as expected.

PR #422, opened 2020-07-14
Release docs

### LINKED ISSUE #378: Add versioned releases
We need to make proper releases that follow semver with change notes, migration guides, etc.

Need to figure out if we need to detach templates from binaries in releases and how to package/release them.

MartinPetkov:
Fix #378


xingao267:
@rtal could you take a look?

RELEASING.md
```diff
@@ -0,0 +1,138 @@
+<!-- markdownlint-configure-file { "MD013": { "line_length": 120 } } -->
+# Releasing
+
+## Overview
+
+This repo contains three types of release artifacts:
+
+* **Go Binaries** for Terraform Engine, Policy Generator, and Terraform Importer.
+* **Templates** for the Terraform Engine ("templates" for short).
+* **Policy templates** for the Policy Generator ("policies" for short).
```
rtal, line 10: Can we link to the main README for explanation of what each of these artifacts are

RELEASING.md
```diff
@@ -0,0 +1,138 @@
...
+
+These three groups of artifacts are not coupled with each other, so they are **released separately**.
+
+### Versioning
+
+Releases are tagged with [Git Annotated Tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging) following [semantic
+versioning](https://semver.org/) (semver):
+
+* `v1.x.x` - MAJOR version, increments on breaking changes.
```
rtal, line 19: Can we potentially list some examples of what we consider breaking changes?
i.e. is pinning to a new version of Terraform considered a breaking change, or are we only considering breaking changes to the resources/functionality entirely contained within our package?

RELEASING.md
```diff
@@ -0,0 +1,138 @@
...
+* `v1.2.x` - MINOR version, increments on backwards-compatible new features.
+* `v1.2.3` - PATCH version, increments on backwards-compatible bug fixes.
+
+The only exception to the above is the `0.x.x` version. As described in semver, this version is for initial development;
+anything may change at any time, and the public API should not be considered stable.
+
+We have the following release tags:
+
+* Ex. `v0.1.2` for binaries
+* Ex. `templates-v0.1.2` for templates.
```
rtal, line 29: Are we versioning all templates as a single unit or will each template have it's own version history? I think we're versioning all templates as a single block, so any single update would revision all templates, but can we clarify that here?

RELEASING.md
```diff
@@ -0,0 +1,138 @@
...
+* Ex. `policies-v0.1.2` for policies.
```
rtal, line 30: Same comment as above, but for policies

RELEASING.md
MartinPetkov: Done.

RELEASING.md
MartinPetkov: All templates are versioned as a single unit, same for policies. I'll try to clarify it.

RELEASING.md
MartinPetkov: Done.

RELEASING.md
MartinPetkov: Done

RELEASING.md
MartinPetkov: I added some examples. Let me know if they're clear, or if I should add more.

rtal: LGTM

RELEASING.md
```diff
@@ -0,0 +1,186 @@
+<!-- markdownlint-configure-file { "MD013": { "line_length": 200 } } -->
+# Releasing
+
+## Overview
+
+This repo contains three types of release artifacts:
+
+* **Go Binaries** for [Terraform Engine](./docs/tfengine), [Policy Generator](./docs/policygen), and [Terraform Importer](./docs/tfimport).
```
umairidris, line 8: Optional: consider including the tags here so it is prominent right at the top

* Go Binaries (tag: `vX.Y.Z`)
*  Templates (tag: `templates-vX.Y.Z`)

RELEASING.md
```diff
@@ -0,0 +1,186 @@
...
+### Caveat about versions
+
+Solutions docs and guides should pin the versions of the binaries, templates and policies to ensure that users will see
+consistent binary behaviour and output files.
+
+However, as cloud APIs and other infrastructure change over time, in practice solutions may stop working even with the
+same versions as before. We cannot guarantee indefinite compatibility, and encourage solutions and guide writers to
+periodically check for correctness and consider updating to use the latest versions.
+
+## Automated releases
```
umairidris, line 74: Should these go in a dev doc? I think this doc should only be for users

RELEASING.md
MartinPetkov: Done.

RELEASING.md
MartinPetkov: I personally don't think so, I see this doc as the one place for release-related information. Also in the past where I've encountered RELEASING.md docs, they have instructions on how to make a release.

PR #426, opened 2020-07-16
Separate build functions

MartinPetkov:
This way they can be individually used in release automation.

### DIFF
--- a/build/build-binaries.sh
+++ b/build/build-binaries.sh
@@ -14,15 +14,13 @@
 
 #!/usr/bin/env bash
 
-# Helper to build release artifacts - binaries, and templates and policies bundles.
+# Helper to build release artifacts - binaries only.
 #
 # Usage (from repo root):
-# ./build/build-release.sh -o . -v v0.1.0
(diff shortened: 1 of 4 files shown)


build/build-binaries.sh
```diff
@@ -0,0 +1,26 @@
...
+# Helper to build release artifacts - binaries only.
+#
+# Usage (from repo root):
+# ./build/build-binaries.sh -o . -v v0.1.0
+
+# Collects args and defines build functions.
+source "$(dirname "${0}")/build-include.sh"
+
+# Build binaries only.
+build_binaries
```
umairidris, line 26: Can we use something like a make file and user can just say what they want to build?

e.g.

make binaries v0.1.0
make templates v.02.0 (append templates- automatically)

MartinPetkov: From what I can tell, you're not really supposed to be calling make with arguments, only targets.

MartinPetkov: I'll see if there's a sensible way to do it with environment variables or something.

umairidris: No wories then, this is fine

PR #427, opened 2020-07-16
Support fetching remote templates

### LINKED ISSUE #233: Fetch template dependencies on the fly
We can borrow the idea that Bazel implements dependencies, e.g. `http_archive`, `http_file`, `git_repository` https://github.com/bazelbuild/bazel/blob/master/WORKSPACE#L378 and we can download the templates on the fly at the beginning of engine and policygen execution.

umairidris:
#233


cmd/tfengine/main.go
```diff
@@ -41,7 +42,8 @@ func main() {
 		log.Fatal("--output_path must be set")
 	}
 
-	if err := tfengine.Run(*configPath, *outputPath, *format); err != nil {
+	opts := &tfengine.Options{Format: *format, CacheDir: *cacheDir}
+	if err := tfengine.Run(*configPath, *outputPath, opts); err != nil {
```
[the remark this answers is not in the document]
umairidris: Options are for optional things that can have a default value. If this was a user facing API I would have implemented these as functional options: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis but we don't need to do that right away as the values are coming from flags.

internal/tfengine/tfengine.go
```diff
@@ -136,25 +141,52 @@ func dumpTemplate(conf *Config, root, outputPath string, ti *templateInfo) error
...
+		}
+		if !filepath.IsAbs(path) {
+			path = filepath.Join(pwd, path)
+		}
+		return path, nil
+
+	}
+	dst := filepath.Join(cacheDir, "templates")
+	root, subdir := getter.SourceDirSubdir(path)
+	c := getter.Client{
```
xingao267, line 178: this is neat

internal/tfengine/tfengine.go
```diff
@@ -136,25 +141,52 @@ func dumpTemplate(conf *Config, root, outputPath string, ti *templateInfo) error
...
 	}
 	return nil
 }
 
+// fetchPath handles fetching remote paths.
+// If the path is local then it will simply expand it and return.
+// Remote paths are fetched to the cache dir.
+// Relative paths are handled from pwd.
+func fetchPath(path, pwd, cacheDir string) (string, error) {
+	if strings.HasPrefix(path, ".") { // Is local path.
```
xingao267, line 165: what if the local path is an absolute path?

internal/tfengine/tfengine.go
```diff
@@ -136,25 +141,52 @@ func dumpTemplate(conf *Config, root, outputPath string, ti *templateInfo) error
...
-			cp = filepath.Join(root, cp)
-		}
 		if err := template.WriteDir(cp, outputPath, data); err != nil {
 			return fmt.Errorf("component %q: %v", cp, err)
 		}
 	}
 	return nil
 }
 
+// fetchPath handles fetching remote paths.
```
xingao267, line 160: move this to a common package so policygen can use it

xingao267: maybe pathutil

cmd/tfengine/main.go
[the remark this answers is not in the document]
MartinPetkov: I see, thanks for the explanation.

internal/tfengine/tfengine.go
MartinPetkov: Maybe Stat the string and check if it's a file path.

umairidris: Absolute paths are not currently supported and involves some care to implement. I based the current logic on https://github.com/hashicorp/terraform/blob/807267d1b56aa6298fecf5edba98c088c2866e83/internal/initwd/getter.go#L177. Terraform does some further checks for absolute paths I think which I can port later on.

umairidris: stat seems like a good idea, I can do that

umairidris: os.Stat isn't working out too well either, I'll revisit this later.

xingao267:
#233 includes both tfengine and policygen. So maybe leave it open after this is merged.

umairidris:
> #233 includes both tfengine and policygen. So maybe leave it open after this is merged.

got it

PR #428, opened 2020-07-16
GitHub Actions workflows for release automation

MartinPetkov:
Docs:
* https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-and-managing-workflow-files-and-runs
* https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token
* https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on
* https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun
* https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
* https://github.com/github/hub#github-actions

### DIFF
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -1,3 +1,17 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 on:
   push:
     tags:
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -22,7 +22,7 @@ jobs:
 
     - name: Build Go binaries
       shell: bash
-      run: ./build/build-binaries.sh -v {{ github.ref }}
+      run: ./build/build-binaries.sh -v ${{ github.ref }}
 
     # From https://github.com/github/hub#github-actions
     - name: Create GitHub Release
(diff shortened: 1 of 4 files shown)


.github/workflows/release-binaries.yaml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Setup Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '^1.14'
+    - run: go version
```
xingao267, line 37: is the - in before run expected?

.github/workflows/release-binaries.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+    - name: Build Go binaries
+      shell: bash
+      run: ./build/build-binaries.sh -v "$(echo '${{ github.ref }}' | sed -e 's|^refs/tags/||')"
+
+    # From https://github.com/github/hub#github-actions
+    - name: Create GitHub Release
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
xingao267, line 46: whose token is this?

MartinPetkov: See https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token:
"GitHub automatically creates a GITHUB_TOKEN secret to use in your workflow. You can use the GITHUB_TOKEN to authenticate in a workflow run.

Before each job begins, GitHub fetches an installation access token for the job. The token expires when the job is finished."

.github/workflows/release-binaries.yaml
MartinPetkov: Yes. This is technically a step without a name, that runs directly on the runner (i.e. not in a Docker container). See https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#choosing-a-runner.

MartinPetkov: And this: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun

MartinPetkov: The action step (`uses: actions/setup-go@v2`) is the equivalent of running docker run without an explicit entrypoint, it uses the action's default command (in this case it's a bunch of commands to download Go and configure the PATH). That's why it doesn't have a "run".

umairidris:
LGTM, will let Xin approve

PR #429, opened 2020-07-16
migrate folders to single pure terraform deployment

umairidris:


templates/tfengine/components/folders/main.tf
```diff
@@ -12,7 +12,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */ -}}
 
-resource "google_folder" "folder" {
+{{range .folders -}}
+resource "google_folder" "{{resourceName . "display_name"}}" {
   display_name = "{{.display_name}}"
-  {{hclField . "parent"}}
+  parent       = "{{get . "parent" $.parent}}"
```
MartinPetkov, line 18: Where does this fallback parent come from? I don't see it in the examples (I assume it would be in the data block).

umairidris: It's coming from folders.hcl

```
data = {
    {{if eq .parent_type "organization"}}
    parent       =  "organizations/{{.parent_id}}"
    {{else}}
    parent       = "folders/{{.parent_id}}"
    {{end}}
  }
```

PR #431, opened 2020-07-17
Finish no terragrunt config changes

umairidris:
Now we just need to port over CICD and test and this should be done.


examples/tfengine/team.hcl
```diff
@@ -341,16 +331,15 @@ template "project_apps" {
...
-        #   ip_range_services_name = "example-services-range"
-        #   master_ipv4_cidr_block = "192.168.0.0/28"
-        # }]
+        gke_clusters = [{
+          name                   = "example-gke-cluster"
+          network_project_id     = "example-prod-networks"
+          network                = "example-network"
+          subnet                 = "example-gke-subnet"
+          ip_range_pods_name     = "example-pods-range"
+           ip_range_services_name = "example-services-range"
```
MartinPetkov, line 340: Extra space

examples/tfengine/team.hcl
```diff
@@ -271,7 +265,7 @@ template "project_data" {
         }]
         iam_members = {
           "roles/cloudsql.client" = [
-            "serviceAccount:$${var.bastion_service_account}",
+            "serviceAccount:bastion@example-prod-networks.iam.gserviceaccount.com",
```
MartinPetkov, line 268: Can you remind me, why are we removing this as a variable?

umairidris: This is actually the default name: https://github.com/terraform-google-modules/terraform-google-bastion-host/blob/master/variables.tf#L130

I can set this more explicitly later so it's clearer where this value is coming from.

PR #433, opened 2020-07-17
support fetch remote policies in policygen

xingao267:
also just use a tmp dir for engine remote template cache

Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/233

### DIFF
--- a/examples/policygen/config.yaml
+++ b/examples/policygen/config.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-template_dir: ../../templates/policygen
+template_dir: "git://github.com/GoogleCloudPlatform/healthcare-data-protection-suite//templates/policygen?ref=policies-v0.1.0"
 
 gcp_org_policies:
   parent_type: organization # One of `organization`, `folder` or `project`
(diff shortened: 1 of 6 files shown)


internal/policygen/load.go
```diff
@@ -132,10 +129,6 @@ func (c *config) init(path string) error {
 		}
 	}
 
-	if !filepath.IsAbs(c.TemplateDir) {
```
MartinPetkov, line 135: Why is this being removed?

xingao267: It's checked in Fetch() now.

examples/policygen/config.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-template_dir: ../../templates/policygen
+template_dir: "git://github.com/GoogleCloudPlatform/healthcare-data-protection-suite//templates/policygen?ref=policies-v0.1.0"
```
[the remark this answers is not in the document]
xingao267: Yes. Both config.yaml and config.hcl are tested, and I made one remote path and one local path. The regen.sh is using the one with local so local changes to the templates will result in new policies being generated.

PR #444, opened 2020-07-21
change to more inclusive language

xingao267:
This is a breaking change. Existing user of the `tf-deletion-whitelist.txt` needs to rename it to `tf-deletion-allowlist.txt`.


xingao267:
> Just FYI that I think this is backwards-incompatible, if anyone happened to be using this feature. They'd have to rename their current `-whitelist.txt` file

Ack. Got it.

PR #449, opened 2020-07-22
fix validate presubmit

xingao267:


xingao267:
TESTED=both terragrunt and terraform only workflow

PR #456, opened 2020-07-23
Simplify gen_check.sh

MartinPetkov:


build/gen_check.sh
```diff
@@ -17,40 +17,22 @@
...
 
-  # Generate in a tmp dir, then diff with examples.
-  # Give the tmpdir a more meaningful name
-  suffix="$(echo "${reference_dir}" | cut -d'/' -f2,4 --output-delimiter='-')"
-  tmp="$(mktemp -d "/tmp/tmp-${suffix}.XXXXXX")"
-  trap "rm -rf '${tmp}'" EXIT INT TERM RETURN ERR
-
-  ${cmd} "${tmp}"
+# Regenerate examples
+./scripts/regen.sh
```
xingao267, line 26: can we regen it to a tmp location instead of polute the source?

MartinPetkov: Done. Added an optional argument to `regen.sh`.

build/gen_check.sh
```diff
@@ -22,11 +22,11 @@ tmp="$(mktemp -d "/tmp/tmp-gen_check.XXXXXX")"
...
 cp -ar './examples/' "${tmp}"
 
-# Regenerate examples
-./scripts/regen.sh
+# Regenerate examples into the tmp dir
+./scripts/regen.sh "${tmp}/examples"
 
 # Check for diffs
-changed="$(diff -qr examples/ ${tmp}/examples/ | grep -v ': README.md')" || true
+changed="$(diff -qr "${tmp}/examples" "./examples" | grep -v ': README.md')" || true
```
xingao267, line 29: seems both the cp and regen write to ${tmp}/examples?

MartinPetkov: Yes, it first copies the directory as it is, then regenerates the files. This makes sure we don't have to ignore a bunch of files or directories for the diff, since everything else should be the same.

PR #457, opened 2020-07-23
Create empty output directories with no files

MartinPetkov:
No particular reason that this should be an error. Someone could use this to set up a directory structure (envs/, modules/) that they'll populate later.

### DIFF
--- a/internal/template/template.go
+++ b/internal/template/template.go
@@ -50,7 +50,7 @@ func WriteDir(inputDir, outputDir string, data map[string]interface{}) error {
 	if len(fs) == 0 {
 		// Just create the output directory with no files.
 		if err := os.MkdirAll(outputDir, 0755); err != nil {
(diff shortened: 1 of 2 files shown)


internal/template/template.go
```diff
@@ -48,7 +48,11 @@ func WriteDir(inputDir, outputDir string, data map[string]interface{}) error {
 		return fmt.Errorf("read dir %q: %v", inputDir, err)
 	}
 	if len(fs) == 0 {
-		return fmt.Errorf("found no files in %q to write", inputDir)
+		// Just create the output directory with no files.
+		if err := os.MkdirAll(outputDir, 0755); err != nil {
+			return fmt.Errorf("mkdir %q: %v", outputDir, err)
```
xingao267, line 53: ("create dir %q: %v", outputDir, err)

xingao267:
LGTM

PR #459, opened 2020-07-23
Multiple actions

umairidris:
terraform apply -auto-approve does not show the plan, which can make things confusing when something goes wrong with one deployment. Show a plan before apply to be safe.

### DIFF
--- a/templates/tfengine/components/cicd/configs/run.sh
+++ b/templates/tfengine/components/cicd/configs/run.sh
@@ -46,6 +46,7 @@ do
     terraform init
     for action in "${ACTIONS[@]}"
     do
+      # Convert action string to array as it can have multiple arguments.
       IFS=' ' read -r -a args <<< "${action}"
       terraform "${args[@]}"
     done
--- a/templates/tfengine/components/cicd/configs/run.sh
+++ b/templates/tfengine/components/cicd/configs/run.sh
@@ -17,20 +17,20 @@
 set -e
 set -x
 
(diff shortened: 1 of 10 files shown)


templates/tfengine/components/cicd/configs/run.sh
```diff
@@ -39,11 +39,14 @@ do
...
-for d in "${DIRS[@]}"
+for mod in "${MODULES[@]}"
 do
-    cd "${ROOT}"/"${d}"
+    cd "${ROOT}"/"${mod}"
     terraform init
-    terraform "${args[@]}"
+    for action in "${ACTIONS[@]}"
+    do
+      IFS=' ' read -r -a args <<< "${action}"
```
MartinPetkov, line 49: Please add a comment, it's not obvious to me what this does or why it's needed.

umairidris: Done. Note the main motivation behind these changes is passing shell check

MartinPetkov: Do we have shellcheck in CI?

umairidris: Not at the moment (I think most of our current scripts would fail it too). I think it's more important for the cloud build scripts to pass shell check as customers would use them. Filed https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/460 to track

Issue #466, opened 2020-07-23
Check go lint issues in presubmit


xingao267:
To catch things like https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/465

xingao267:
Use https://github.com/golang/lint in presubmit

PR #467, opened 2020-07-23
add golint in ci

xingao267:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/466


internal/policygen/policygen.go
```diff
@@ -43,7 +43,6 @@ type RunArgs struct {
 	OutputPath string
 }
 
-// Run executes main policygen logic.
```
umairidris, line 46: This is a public func. It doesn't need a comment?

internal/runner/runner.go
```diff
@@ -75,15 +75,18 @@ func (d *Default) CmdCombinedOutput(cmd *exec.Cmd) ([]byte, error) {
 // Dry is the Runner that only prints commands and does not execute them.
 type Dry struct{}
 
+// CmdRun prints the command.
```
umairidris, line 78: Runs?

internal/runner/runner.go
```diff
@@ -75,15 +75,18 @@ func (d *Default) CmdCombinedOutput(cmd *exec.Cmd) ([]byte, error) {
...
 func (*Dry) CmdRun(cmd *exec.Cmd) error {
 	log.Printf("%v", cmd.String())
 	return nil
 }
 
+// CmdOutput prints the command.
```
umairidris, line 84: Runs the command and captures stdout

internal/policygen/policygen.go
xingao267: it needs. you caught this right in the meantime of me testing the CI should fail this case.

internal/runner/runner.go
```diff
@@ -75,15 +75,18 @@ func (d *Default) CmdCombinedOutput(cmd *exec.Cmd) ([]byte, error) {
...
 func (d *Dry) CmdOutput(cmd *exec.Cmd) ([]byte, error) {
 	return []byte(cmd.String()), d.CmdRun(cmd)
 }
 
+// CmdCombinedOutput prints the command.
```
umairidris, line 89: Runs the command and captures stdout and stderr.

internal/policygen/policygen.go
xingao267: It's reverted in the next commit.

internal/runner/runner.go
xingao267: It's dry run, so it's just printing the commands.

internal/runner/runner.go
xingao267: ditto

internal/runner/runner.go
xingao267: ditto

xingao267:
@googlebot rescan

Issue #476, opened 2020-07-24
CICD field renames


umairidris:
```
triggers = {
  validate = {}
  plan =  {}
  apply = {}
}
```

Also rename disabled to run_on_push and default to true.

Release as part of 0.2.

PR #484, opened 2020-07-27
improve cicd trigger block

xingao267:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/476

### DIFF
--- a/templates/tfengine/recipes/devops.hcl
+++ b/templates/tfengine/recipes/devops.hcl
@@ -95,6 +95,7 @@ schema = {
       additionalProperties = false
       required = [
         "branch_regex",
+        "triggers",
       ]
       properties = {
         github = {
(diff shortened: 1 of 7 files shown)


templates/tfengine/recipes/devops.hcl
```diff
@@ -231,7 +238,7 @@ template "root" {
 {{end}}
 
 # At least one trigger is specified.
-{{if and (has . "cicd") (or (has .cicd "validate_trigger") (has .cicd "plan_trigger") (has .cicd "apply_trigger"))}}
+{{if and (has . "cicd") (or (has .cicd.triggers "validate") (has .cicd.triggers "plan") (has .cicd.triggers "apply"))}}
```
MartinPetkov, line 241: Nit: Would look cleaner with a `hasOne` function, so like:
```
{{if and (has . "cicd") (hasOne .cicd.triggers "validate" "plan" "apply")}}
```

xingao267: created https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/485 and added TODO

PR #493, opened 2020-07-28
Increase the hourly PR limit

MartinPetkov:


MartinPetkov:
This is actually probably not the right solutions. I'll see if I can configure Renovate Bot to update all versions at the same time.

PR #508, opened 2020-07-29
update cicd state path

xingao267:


xingao267:
Close in favor of https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/509

Issue #438, opened 2020-07-20
Support writing out individual files from components


MartinPetkov:
Example:
```
template "my_template" {
  component_path = "../components/state.tf"
  output_path = "./mystate.tf"
}
```
Result: creates a files `./mystate.tf`, ignoring the rest of the files in that component.

PR #511, opened 2020-07-30
Supporting outputting individual component files

MartinPetkov:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/438

### DIFF
--- a/internal/tfengine/tfengine.go
+++ b/internal/tfengine/tfengine.go
@@ -156,11 +156,16 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
 			return err
 		}
 
+		write := template.WriteDir
+
 		// If the component path is a single file,
 		// treat the output_path as a file name and only write that out.
 		// Otherwise, treat as directory path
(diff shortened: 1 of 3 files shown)


internal/tfengine/tfengine.go
```diff
@@ -150,11 +150,21 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
...
 		if err != nil {
 			return err
 		}
-		if err := template.WriteDir(cp, outputPath, data); err != nil {
+
+		// If the component path is a single file,
+		// treat the output_path as a file name and only write that out.
+		// Otherwise, treat as directory path
+		write := template.WriteDir
+		if pathutil.IsFile(cp) {
```
umairidris, line 163: I am guessing the os.Stat should not raise an error so I would just do it here and return the err if it isn't nil.

MartinPetkov: I was thinking that if it raises an error, that's still a "no". But you're right, it should probably not be swallowed. I'll change it.

PR #512, opened 2020-07-30
fix secret locations

xingao267:
https://www.terraform.io/docs/providers/google/r/secret_manager_secret.html


examples/tfengine/team.hcl
```diff
@@ -118,6 +118,10 @@ EOF
       secrets = [
         {
           secret_id = "manual-sql-db-user"
+          locations = [
```
umairidris, line 121: To be consistent with others can we call this secret_locations?

xingao267: done

PR #514, opened 2020-07-30
Fix blank lines

### LINKED ISSUE #356: Format generated files

umairidris:
Move - to the end of the {{ }} blocks. Normally this would cause the output to be messed up but terraform fmt cleans it up quite nicely.

I chose to do this over a helper that replaces things with a regex because this also helps in making sure there is at least one blank line between major blocks.

Fix #356


examples/tfengine/generated/org_foundation/monitor/main.tf
```diff
@@ -39,8 +38,6 @@ module "project" {
   activate_apis = [
   ]
 }
-
```
xingao267, line 42: can we also leave one empty liine in cases like this?

examples/tfengine/generated/team/example-prod-networks/main.tf
```diff
@@ -45,10 +44,10 @@ module "project" {
     "sqladmin.googleapis.com",
   ]
 }
-
```
xingao267, line 48: and this

PR #518, opened 2020-07-31
Fix renovate bot

MartinPetkov:
Use the regex manager to match all "source/version" pairs in .tf files.

This is based on the suggestion [here](https://github.com/renovatebot/config-help/issues/782#issuecomment-665427224), with some testing based on [the self-hosted guide](https://github.com/renovatebot/renovate/blob/master/docs/development/self-hosting.md) and using [my fork](https://github.com/MartinPetkov/healthcare-data-protection-suite/blob/master/renovate.json)


xingao267:
Thanks a lot for looking into this!

PR #520, opened 2020-07-31
Correct regex match field according to docs

MartinPetkov:
Changes made since last time:
* Target the "regex" manager in the packageRules rule.
* Put `fileMatch` at the top level, where it should be.
* Put `regexManagers` at the top level, where it should be.
* Adjust the regex to handle both deps with submodules (`//`) and without.

The regex was tested against these files via [regex101](https://regex101.com/):
* https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/examples/tfengine/generated/team/example-prod-apps/main.tf
* https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/components/resources/gke_clusters/main.tf

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,13 +5,15 @@
   "ignorePaths": [],
   "packageRules": [
     {
-      "managers": ["terraform"],
-      "groupName": "terraform",
-      "fileMatch": [
-        "\\.tf$"
-      ],
+      "managers": ["regex", "terraform"],
+      "groupName": "terraform"
+    }
+  ],
+  "fileMatch": ["\\.tf$"],
+  "regexManagers": [
+    {
       "matchStrings": [
-        "source\\s+=\\s+\"(?<depName>.+?)\"\\s+version\\s+=\\s+\"(?<currentValue>.+?)\""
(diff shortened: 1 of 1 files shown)


MartinPetkov:
The way I tested:
* Commit the `renovate.json` config in [my fork](https://github.com/MartinPetkov/healthcare-data-protection-suite/blob/master/renovate.json) on the `master` branch
* Install renovate: `npm i -g renovate`
* Write a config.js:

    ```js
    module.exports = {
      endpoint: 'https://api.github.com',
      token: '<redacted>',
      platform: 'github',
      logFileLevel: 'warn',
      logLevel: 'debug',
      logFile: './renovate.log',
      onboarding: true,
      dryRun: true,
      onboardingConfig: {
        extends: ['config:base'],
      }
      repositories: ['MartinPetkov/healthcare-data-protection-suite'],
    };
    ```
* Run renovate locally: `renovate --dry-run --include-forks`
* Look for warnings (there shouldn't be any) and for the results under `packageFiles with updates` under the "terraform" manager. There should be results for files under `templates/` as well as under `examples/`.

Issue #532, opened 2020-08-05
Remove org policies from policygen


umairidris:
We should just support this in the engine so that it can be deployed as part of a full e2e run (e.g. with state etc set).

Release as part of 0.2

PR #533, opened 2020-08-05
move org policy to tfengine

xingao267:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/532


docs/tfengine/README.md
```diff
@@ -14,7 +14,7 @@ used to generate
 specific to your Google Cloud organization and structure.
 
 Users typically pick from a set of
-[recipes](../../templates/tfengine/schemas) which implement a template for one
+[recipes](../../templates/tfengine/recipes) which implement a template for one
```
umairidris, line 17: Ah I remember now what I wanted to do. I was going to point this to docs/tfengine/schemas. WDYT is a better link?

templates/tfengine/recipes/org_policies.hcl
```diff
@@ -12,12 +12,91 @@
...
+  }
+}
 
 template "deployment" {
   recipe_path = "./deployment.hcl"
 }
 
 template "org_policies" {
-  component_path = "../../policygen/org_policies"
+  component_path = "../components/security/org_policies"
```
umairidris, line 101: Do we have anything else under security/ ?

docs/tfengine/README.md
xingao267: You have another link in the next sentences that is refering to the md folder.

templates/tfengine/recipes/org_policies.hcl
xingao267: not right now. I can move org_policies one level up

umairidris: I would vote for keeping it one level higher but if you think we will expand this its fine to leave as is

xingao267: Done

Issue #544, opened 2020-08-10
Add flag to generate specific blocks in an Engine config


MartinPetkov:
For example, `--config_path 'examples/tfengine/team.hcl' --templates devops,cicd` would generate only those two parts of the [`team.hcl`](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/examples/tfengine/team.hcl) config.

PR #545, opened 2020-08-10
Add --templates flag to Engine

MartinPetkov:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/544

### DIFF
--- a/cmd/tfengine/main.go
+++ b/cmd/tfengine/main.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"strings"
 
 	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/tfengine"
 )
@@ -31,6 +32,7 @@ var (
 	configPath = flag.String("config_path", "", "Path to config file")
 	outputPath = flag.String("output_path", "", "Path to directory dump output")
 	format     = flag.Bool("format", true, "Whether to format generated files.")
+	templates  = flag.String("templates", "", "Comma-separated list of templates to generate. Leave empty for all.")
 )
 
 func main() {
@@ -49,7 +51,19 @@ func main() {
 	}
 	defer os.RemoveAll(cacheDir)
 
-	opts := &tfengine.Options{Format: *format, CacheDir: cacheDir}
+	templatesMap := make(map[string]bool)
+	for _, t := range strings.Split(*templates, ",") {
+		t = strings.TrimSpace(t)
+		if len(t) > 0 {
+			templatesMap[t] = true
+		}
+	}
+
+	opts := &tfengine.Options{
+		Format:    *format,
+		CacheDir:  cacheDir,
+		Templates: templatesMap,
+	}
 	if err := tfengine.Run(*configPath, *outputPath, opts); err != nil {
 		log.Fatal(err)
 	}
--- a/cmd/tfengine/main.go
+++ b/cmd/tfengine/main.go
@@ -51,18 +51,18 @@ func main() {
 	}
 	defer os.RemoveAll(cacheDir)
 
-	templatesMap := make(map[string]bool)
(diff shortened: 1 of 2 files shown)


internal/tfengine/tfengine.go
```diff
@@ -33,8 +33,9 @@ import (
 
 // Options is the options for tfengine execution.
 type Options struct {
-	Format   bool
-	CacheDir string
+	Format    bool
+	CacheDir  string
+	Templates map[string]bool
```
umairidris, line 38: Naming this Templates is a bit confusing. Maybe something like WantedTemplates or something? 

Also add a comment above this field to describe its purpose and that a len of 0 means get all templates.

internal/tfengine/tfengine.go
```diff
@@ -80,8 +81,13 @@ func Run(confPath, outPath string, opts *Options) error {
 	return copy.Copy(tmpDir, outPath)
 }
 
-func dump(conf *Config, pwd, cacheDir, outputPath string) error {
+func dump(conf *Config, pwd, cacheDir, outputPath string, templates map[string]bool) error {
 	for _, ti := range conf.Templates {
+		// If a templates filter was provided, check against it.
+		if _, ok := templates[ti.Name]; len(templates) > 0 && !ok {
```
umairidris, line 87: optional: if you pop the key off the map then you can check for whether the list is non-empty at the end and detect the cases where an invalid key was passed in (e.g. to catch typos etc).

cmd/tfengine/main.go
```diff
@@ -49,7 +51,19 @@ func main() {
 	}
 	defer os.RemoveAll(cacheDir)
 
-	opts := &tfengine.Options{Format: *format, CacheDir: cacheDir}
+	templatesMap := make(map[string]bool)
+	for _, t := range strings.Split(*templates, ",") {
+		t = strings.TrimSpace(t)
+		if len(t) > 0 {
```
xingao267, line 57: do we need this if we have trimmed spaces?

MartinPetkov: Yes. Otherwise, if you provide no flag or `--templates ,,`, it ends up mapping `""` (empty string) to `true`.

internal/tfengine/tfengine.go
MartinPetkov: Done.

internal/tfengine/tfengine.go
MartinPetkov: Added. Opted not to pop the key, to avoid deleting the data and possibly causing issues in the future.

umairidris: You're still mutating the input so I would prefer popping as that would be simpler, but your call.

internal/tfengine/tfengine.go
```diff
@@ -80,12 +84,38 @@ func Run(confPath, outPath string, opts *Options) error {
...
+		}
+
 		if err := dumpTemplate(conf, pwd, cacheDir, outputPath, ti); err != nil {
 			return fmt.Errorf("template %q: %v", ti.Name, err)
 		}
 	}
+
+	// Report on any templates that were supposed to be generated but weren't.
+	// This most likely indicates a misspelling.
+	var wrongTemplates []string
```
umairidris, line 107: missingTemplates ?

internal/tfengine/tfengine.go
umairidris: Another option could have been to make a copy of the map and pass that in, though I am not sure if that makes things clearer.

MartinPetkov: It gets reset back to the way it was at the end, so the function call is a no-op for that map.

internal/tfengine/tfengine.go
MartinPetkov: Done.

PR #552, opened 2020-08-14
Add more information to loadConfig errors

MartinPetkov:


internal/tfengine/config.go
```diff
@@ -112,19 +112,19 @@ func (c *Config) validate() error {
 func loadConfig(path string, data map[string]interface{}) (*Config, error) {
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("ReadFile %v: %v", path, err)
```
umairidris, line 115: write user facing errors as descriptions rather than func names

e.g.

read file:

internal/tfengine/config.go
```diff
@@ -112,19 +112,19 @@ func (c *Config) validate() error {
...
 	}
 	buf, err := template.WriteBuffer(string(b), data)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("WriteBuffer %v: %v", string(b), err)
```
umairidris, line 119: parsing template:

internal/tfengine/config.go
```diff
@@ -112,19 +112,19 @@ func (c *Config) validate() error {
...
 	}
 
 	// Convert yaml to json so hcl decoder can parse it.
 	cj := buf.Bytes()
 	if filepath.Ext(path) == ".yaml" {
 		cj, err = yaml.YAMLToJSON(cj)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("YAMLToJSON %v: %v", string(cj), err)
```
umairidris, line 127: also error strings should be lower case so they can be chained

PR #554, opened 2020-08-17
project factory v9

umairidris:


templates/tfengine/components/project/main.tf
```diff
@@ -9,10 +9,16 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */ -}}
 
+{{$source := "terraform-google-modules/project-factory/google" -}}
+{{if has . "shared_vpc_attachment" -}}
```
xingao267, line 13: do we need to account for the case when `is_shared_vpc_host = true`?

umairidris: I don't think so. This is only needed by service projects that are hosting GKE and dataproc instances.

Issue #528, opened 2020-08-05
support IAM bindings in policygen


xingao267:
Start with just merging IAM member and bindings to be less restrictive. The policy generated currently is already quite restrictive.

PR #559, opened 2020-08-24
support google_*_iam_bindings in policygen

xingao267:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/528


internal/policygen/iam.go
```diff
@@ -95,54 +95,120 @@ func generateIAMPolicies(rn runner.Runner, resources []*states.Resource, outputP
...
+		return nil, fmt.Errorf("get resource instances for type %q: %v", resourceType, err)
+	}
 
-			// Init the roleBindings map if it didn't exist.
-			if _, ok := bindings[key]; !ok {
-				bindings[key] = make(roleBindings)
-			}
+	for _, ins := range instances {
+		if err := validateMandatoryStringFields(ins, []string{idField, "role", "member"}); err != nil {
+			return nil, err
```
MartinPetkov, line 146: Maybe add "fmt.Errorf()" to indicate this error comes from missing fields in the instance.

internal/policygen/iam.go
```diff
@@ -95,54 +95,120 @@ func generateIAMPolicies(rn runner.Runner, resources []*states.Resource, outputP
...
+		}
+
+		id, err := normalizeID(rn, rootType, ins[idField].(string)) // Type checked in validate function.
+		if err != nil {
+			return nil, err
```
MartinPetkov, line 151: Same here, add fmt.Errorf() to give context to the error.

internal/policygen/iam.go
```diff
@@ -154,6 +220,39 @@ func validate(instance map[string]interface{}, mandatoryFields []string) error {
 	return nil
 }
 
+// validateMandatoryLists checks the presence of mandatory fields and assert []interface{} type.
+func validateMandatoryStringLists(instance map[string]interface{}, mandatoryFields []string) error {
+	for _, k := range mandatoryFields {
+		if _, ok := instance[k]; !ok {
```
MartinPetkov, line 226: I would rewrite this something like the following:

```go
field, ok := instance[k]
if !ok {
	return fmt.Errorf("mandatory field %q missing from instance: %v", k, instance)
}

lst, ok := field.([]interface{})
if !ok {
	return fmt.Errorf("value for %q should be a []interface{}, got %T", k, field)
}

for _, s := range lst {
	if _, ok := s.(string); !ok {
		return fmt.Errorf("%q should be a string, got %T", s, s)
	}
}
```

internal/policygen/iam.go
```diff
@@ -154,6 +220,39 @@ func validate(instance map[string]interface{}, mandatoryFields []string) error {
...
+
+	}
+	return nil
+}
+
+func normalizeID(rn runner.Runner, t, id string) (string, error) {
+	var err error
+	// For projects, the ID in the state is the project ID, but we need project number in policies.
+	if t == "project" {
+		if id, err = projectNumber(rn, id); err != nil {
```
MartinPetkov, line 246: I think it's unexpected to reuse the input arg as the output arg like this. I would instead just make the two output args named and use those. Something like

```go
func normalizeID(rn runner.Runner, t, id string) (nid string, err error) {
	// For projects, the ID in the state is the project ID, but we need project number in policies.
	if t == "project" {
		if nid, err = projectNumber(rn, id); err != nil {
			return "", err
		}
	} else if t == "folder" {
		// For folders, the ID can be either {folder_id} or folders/{folder_id}. Remove the 'folders/' prefix if exists.
		nid = strings.TrimPrefix(id, "folders/")
	}
	return nid, nil
}
```

internal/policygen/iam.go
```diff
@@ -95,54 +95,120 @@ func generateIAMPolicies(rn runner.Runner, resources []*states.Resource, outputP
...
 		"folder":       "folder",
 		"organization": "org_id",
 	}
 
-	for t, idField := range typeToIDField {
-		// TODO(https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/152): Support google_*_iam_bindings.
-		resourceType := fmt.Sprintf("google_%s_iam_member", t)
-		instances, err := terraform.GetInstancesForType(resources, resourceType)
+	for rootType, idField := range typeToIDField {
+		IAMMembers, err := members(rn, resources, rootType, idField)
```
umairidris, line 107: iamMembers

internal/policygen/iam.go
```diff
@@ -95,54 +95,120 @@ func generateIAMPolicies(rn runner.Runner, resources []*states.Resource, outputP
...
 		if err != nil {
-			return nil, fmt.Errorf("get resource instances for type %q: %v", resourceType, err)
+			return nil, err
 		}
 
-		for _, ins := range instances {
-			if err := validate(ins, []string{idField, "role", "member"}); err != nil {
-				return nil, err
-			}
+		IAMBindings, err := bindings(rn, resources, rootType, idField)
```
umairidris, line 112: iamBindings

internal/policygen/iam.go
xingao267: It's already indicated in the `validateMandatoryStringFields` function that it's a mandatory field missing or wrong value type. It's a bit verbose to repeat the same thing here.

internal/policygen/iam.go
xingao267: Done

internal/policygen/iam.go
xingao267: Done

Issue #470, opened 2020-07-24
Support reading from a list of remote state buckets.


xingao267:
Now that we have multiple devops project, states for an org or a folder can be located in multiple state buckets in multiple devops project.

PR #567, opened 2020-08-31
[Breaking change] policygen: support multiple state paths

xingao267:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/470


internal/policygen/policygen.go
```diff
@@ -38,7 +40,7 @@ const (
 // RunArgs is the struct representing the arguments passed to Run().
 type RunArgs struct {
 	ConfigPath string
-	StatePath  string
+	StatePaths string
```
umairidris, line 43: why not make this a list of string and do the split in the caller?

umairidris:
When are we going to do the next major release? Maybe we should keep state_path around till then and just append it to the state paths list?

internal/policygen/policygen.go
xingao267: I'm ok either way. I feel using a comma separated list makes the command line a bit more compact and is what `--templates` is doing tin the tfengine command line.

xingao267:
> When are we going to do the next major release? Maybe we should keep state_path around till then and just append it to the state paths list? Then we can remove it when there is more a necessity for a breaking change.

We are releasing first week of each month, so technically this week or next week. I can hold this PR and merge until we are making a release later this week.

internal/policygen/policygen.go
umairidris: tfengine seems to do it in the main as well: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/cmd/tfengine/main.go#L67

I would vote for doing this in main. Seems odd to have a  StatePaths (plural) that's a string (singular). list of strings is more explicit.

xingao267: Done. Thought you were referring to custom flag. Moved to main.go

PR #568, opened 2020-08-31
Add label support for more engine resources

### LINKED ISSUE #530: Add support for setting labels
It is currently not possible to easily set labels on resources.

We should support:

top level labels block that gets set on all resources
per resource additional labels

The two lists should get merged.

We handle discrepancies between CFT label/field names.

umairidris:
Fix #530


examples/tfengine/team.hcl
```diff
@@ -32,7 +32,7 @@ data = {
   secret_locations    = ["us-central1"]
 
   labels = {
-    "env" = "prod"
+    env = "prod"
```
xingao267, line 35: add one example to demonstrate overriding a label in an individual resource?

umairidris: Added labels to all supported resources.

xingao267 (approved):

PR #578, opened 2020-09-09
Separate the Engine integration tests

MartinPetkov:
Currently all tests in tfengine_test.go are non-hermetic and closer to integration tests, since they aren't Go-only tests. They expect the "terraform" binary to be installed and access GitHub Releases.

Rename this file so that if unit tests are added in the future, there's a clear separation.


umairidris:
Is there supposed to be a renamed file?

MartinPetkov:
> Is there supposed to be a renamed file?

Yes. I just added it, it was ignored by .gitignore before.

umairidris:
I think just calling it integration_test.go for now is fine. We might have a 'full' integration test later on though that actually applies things too, so we may need a way to separate this 'semi' integration test from that.

umairidris (approved):

MartinPetkov:
> I think just calling it integration_test.go for now is fine. We might have a 'full' integration test later on though that actually applies things too, so we may need a way to separate this 'semi' integration test from that.

Done.

PR #580, opened 2020-09-10
Make policygen forseti library a runtime arg

MartinPetkov:
This lets us make the tests hermetic and not have to reach out to the internet.


internal/policygen/policygen_test.go
```diff
@@ -208,9 +208,10 @@ func TestExamples(t *testing.T) {
 		defer os.RemoveAll(tmp)
 
 		args := &RunArgs{
-			ConfigPath: tc.configPath,
-			StatePaths: tc.statePaths,
-			OutputPath: tmp,
+			ConfigPath:  tc.configPath,
+			StatePaths:  tc.statePaths,
+			OutputPath:  tmp,
+			ForsetiPath: "./testdata/policy-library",
```
xingao267, line 214: I'm not sure about checking-in a fixed version of policy-library in our repo, which would need to be updated regularly.  Can we keep this as `git://github.com/forseti-security/policy-library?ref=master` and when importing, string replace `git://github.com/forseti-security/policy-library?ref=master` to an Google internal repo path?

MartinPetkov:
I've actually found a different way to do what I need when importing, so I'm closing this PR.

PR #583, opened 2020-09-15
Add basic version constraint checking to the engine and policygen

MartinPetkov:
Leverages the https://github.com/hashicorp/go-version library.

### DIFF
--- a/cmd/policygen/main.go
+++ b/cmd/policygen/main.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"strings"
 
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd"
 	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/policygen"
 	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/runner"
 )
@@ -36,6 +37,10 @@ var (
 
 func main() {
 	flag.Parse()
+
+	if *cmd.FlagVersion {
+		cmd.ShowVersion()
+	}
 
 	if *configPath == "" {
 		log.Fatal("--config_path must be set")
--- a/cmd/policygen/main.go
+++ b/cmd/policygen/main.go
@@ -21,6 +21,7 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"log"
 	"strings"
 
@@ -37,18 +38,25 @@ var (
 )
 
 func main() {
+	if err := run(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run() error {
 	flag.Parse()
 
 	if *showVersion {
-		cmd.ShowVersion()
+		fmt.Println(cmd.Version)
+		return nil
 	}
 
 	if *configPath == "" {
-		log.Fatal("--config_path must be set")
+		return fmt.Errorf("--config_path must be set")
 	}
 
 	if *outputPath == "" {
-		log.Fatal("--output_path must be set")
+		return fmt.Errorf("--output_path must be set")
 	}
 
 	var statePathsList []string
@@ -68,6 +76,8 @@ func main() {
 
 	rn := &runner.Default{Quiet: true}
 	if err := policygen.Run(context.Background(), rn, args); err != nil {
-		log.Fatalf("Failed to generate policies: %v", err)
+		return fmt.Errorf("Failed to generate policies: %v", err)
 	}
+
+	return nil
 }
--- a/cmd/policygen/main.go
+++ b/cmd/policygen/main.go
@@ -30,15 +30,16 @@ import (
 )
 
 var (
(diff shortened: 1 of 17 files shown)


internal/version/version.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd"
+	"github.com/hashicorp/go-version"
+)
+
+// Compatible returns whether the given string version constraint is compatible with the binary version.
+// Compatibility is determined by github.com/hashicorp/go-version.
+// It returns an error if the constraint string or binary version cannot be interpreted.
+func Compatible(constraint string) (bool, error) {
```
umairidris, line 28: IsCompatible

internal/version/version.go
```diff
@@ -0,0 +1,52 @@
...
+	// If either version is unspecified, it's compatible with all.
+	// Think of it as not setting a compatibility restriction at all.
+	if constraint == "" || cmd.Version == "" {
+		return true, nil
+	}
+
+	// If the bin version is the special "latest", it always succeeds.
+	// By definition there isn't a newer version.
+	if cmd.Version == "latest" {
```
umairidris, line 37: When is latest used?

cmd/version.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+var (
+	// Version is set at compile time with -ldflags "-X <package>/cmd.Version=x.y.z"
+	// Example:
+	//   go run -ldflags "-X github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd.Version=1.2.3" ./cmd/tfengine -version
+	Version = ""
+
+	// FlagVersion indicates a binary should just show the version and exit.
+	FlagVersion = flag.Bool("version", false, "show version and exit")
```
umairidris, line 32: Flags should not be defined in packages, only the main binary. I would just duplicate this across all binaries.

It might also be clearer to say --show_version. If I see --version I would assume it means it only shows version and then exits.

umairidris (approved):

cmd/version.go
MartinPetkov: I moved the flag as suggested.

I'd like to keep it as just -version though, as it _does_ only show the version and exit.

umairidris: Ah, I just saw that.

Could we just do

if *FlagVersion {
  fmt.Println(cmd.Version)
  return nil
}

MartinPetkov: Done. I also changed the binaries to call `run()` in their `main()` and only call `log.Fatal(err)` in one place.

cmd/policygen/main.go
```diff
@@ -62,6 +76,8 @@ func main() {
 
 	rn := &runner.Default{Quiet: true}
 	if err := policygen.Run(context.Background(), rn, args); err != nil {
-		log.Fatalf("Failed to generate policies: %v", err)
+		return fmt.Errorf("Failed to generate policies: %v", err)
```
umairidris: lower case

cmd/version.go
umairidris: Ah right

p.s. you could also have done an empty return if it was in main() (since there is no error returned)

cmd/policygen/main.go
MartinPetkov: It is now. You might be looking at an older copy.

internal/policygen/schema.go
```diff
@@ -22,6 +22,11 @@ additionalProperties = false
 required = ["template_dir"]
 
 properties = {
+  version = {
+    description = "Optional constraint on the binary version required for this config."
```
xingao267, line 26: Given an example of the syntax of the constraint?

MartinPetkov: Good point. The syntax is [here](https://www.terraform.io/docs/configuration/version-constraints.html), added links to it.

cmd/version.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cmd provides access to the embedded version of the binaries.
+// Taken from https://github.com/AgentZombie/go-embed-version.
+package cmd
```
xingao267, line 17: Is there other place to put it? Putting it under cmd package seems a little weird, as it's the top level dir for the 3 binaries.

MartinPetkov: That was the suggested location from that link. It kinds of does feel like the right place to me, since it's a version that only applies to the binaries, all of which are under cmd/.

Would you rather something like cmd/version/?

xingao267: makes sense to me now

xingao267 (approved):

PR #604, opened 2020-10-12
fix(renovate): match .tf files only for terraform manager

rarkins:


MartinPetkov:
Does this fix any bug?

rarkins:
It fixes an incorrect config, but it's not yet detected automatically by Renovate. Once we add detection then this repo would be blocked until the Pr is merged. 

You are overwriting every single manager's default fileMatch with a terraform pattern, so currently it's asking Renovate to scan .tf files for every possible manager so it's doing a massive amount of wasted work attempting to parse them as npm package.json, maven. Pom, etc.

rarkins:
It also means all non-Terraform files are essentially ignored

MartinPetkov:
Okay, that makes sense. Will approve.

MartinPetkov (approved):

MartinPetkov:
I think this PR might have been responsible for why the gen_check is failing in github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/610. It only updated one of the files, and not the other one.

I'm going to revert it, as things were working before.

PR #612, opened 2020-10-16
Revert "fix(renovate): match .tf files only for terraform manager"

MartinPetkov:
Reverts GoogleCloudPlatform/healthcare-data-protection-suite#604

I think this PR might have been responsible for why the gen_check is failing in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/610. It only updated one of the files, and not the other one.

I'm going to revert it, as things were working before.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,9 +9,7 @@
       "groupName": "terraform"
     }
   ],
-  "terraform": {
-    "fileMatch": ["\\.tf$"]
-  },
+  "fileMatch": ["\\.tf$"],
   "regexManagers": [
     {
       "matchStrings": [


rarkins:
No, this is a mistake. You should not merge this PR. Please create an issue in https://github.com/renovatebot/config-help to describe what problem you are trying to resolve and we'll help you promptly

MartinPetkov:
> No, this is a mistake. You should not merge this PR. Please create an issue in https://github.com/renovatebot/config-help to describe what problem you are trying to resolve and we'll help you promptly

I will open a bug, but in the meantime it's better to revert to a known-good state while investigating the fix.

MartinPetkov:
> No, this is a mistake. You should not merge this PR. Please create an issue in https://github.com/renovatebot/config-help to describe what problem you are trying to resolve and we'll help you promptly

I opened https://github.com/renovatebot/config-help/issues/927. But in the meantime I would still strongly prefer to revert to known-good state until we've tested and confirmed a fix, otherwise this will block PRs to our repo from renovate-bot (they can be fixed, but it's annoying to have to do it every time).

PR #613, opened 2020-10-17
Migrate to CFT logging modules

umairidris:
I have upstreamed features from DPT into the CFT logging modules so they mostly match our feature set. They also have new features to set things like bucket locks which are useful for hipaa alignment.

### DIFF
--- a/templates/tfengine/components/audit/main.tf
+++ b/templates/tfengine/components/audit/main.tf
@@ -77,6 +77,9 @@ module "storage_export" {
   include_children       = true
 }
 
+// 6 years minimum audit log retention is required for HIPAA alignment.
+// Thus, we lock retention policy to be at least 6 years
+// and set the actual expiry to be greater than this amount (7 years).
 module "storage_destination" {
   source  = "terraform-google-modules/log-export/google//modules/storage"
   version = "~> 5.0.0"
--- a/templates/tfengine/components/audit/main.tf
+++ b/templates/tfengine/components/audit/main.tf
@@ -86,7 +86,7 @@ module "storage_destination" {
   location                 = "{{.storage_location}}"
   log_sink_writer_identity = "${module.storage_export.writer_identity}"
   storage_class            = "COLDLINE"
-  expiration_days          = 7*365
+  expiration_days          = 7 * 365
   retention_policy = {
     is_locked             = true
     retention_period_days = 6 * 365
--- a/templates/tfengine/components/audit/main.tf
+++ b/templates/tfengine/components/audit/main.tf
@@ -59,6 +59,7 @@ module "bigquery_destination" {
 
   dataset_name             = "{{.logs_bigquery_dataset.dataset_id}}"
   project_id               = module.project.project_id
+  location                 = "{{.bigquery_location}}"
   log_sink_writer_identity = "${module.bigquery_export.writer_identity}"
   expiration_days          = 365
 }
@@ -82,6 +83,7 @@ module "storage_destination" {
 
   storage_bucket_name      = "{{.logs_storage_bucket.name}}"
   project_id               = module.project.project_id
+  location                 = "{{.storage_location}}"
   log_sink_writer_identity = "${module.storage_export.writer_identity}"
   storage_class            = "COLDLINE"
   expiration_days          = 7*365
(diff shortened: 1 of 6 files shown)


MartinPetkov (approved):

templates/tfengine/components/audit/main.tf
```diff
@@ -40,84 +40,67 @@ resource "google_{{.parent_type}}_iam_audit_config" "config" {
...
-
-  dataset_id                  = "{{.logs_bigquery_dataset.dataset_id}}"
-  project_id                  = module.project.project_id
-  location                    = "{{.bigquery_location}}"
-  default_table_expiration_ms = 365 * 8.64 * pow(10, 7) # 365 days
-  access = [
-    {
-      role          = "roles/bigquery.dataOwner",
-      special_group = "projectOwners"
-    },
```
xingao267, line 64: do we still grant this permissions separately?

templates/tfengine/components/audit/main.tf
```diff
@@ -40,84 +40,67 @@ resource "google_{{.parent_type}}_iam_audit_config" "config" {
...
+  location                 = "{{.bigquery_location}}"
+  log_sink_writer_identity = "${module.bigquery_export.writer_identity}"
+  expiration_days          = 365
 }
 
-resource "google_project_iam_member" "bigquery_sink_member" {
-  project = module.bigquery_destination.bigquery_dataset.project
-  role    = "roles/bigquery.dataEditor"
-  member  = google_logging_{{.parent_type}}_sink.bigquery_audit_logs_sink.writer_identity
-}
```
xingao267, line 76: do we need this or it's now part of the log-export module?

templates/tfengine/components/audit/main.tf
```diff
@@ -40,84 +40,67 @@ resource "google_{{.parent_type}}_iam_audit_config" "config" {
...
+
+  storage_bucket_name      = "{{.logs_storage_bucket.name}}"
+  project_id               = module.project.project_id
+  location                 = "{{.storage_location}}"
+  log_sink_writer_identity = "${module.storage_export.writer_identity}"
+  storage_class            = "COLDLINE"
+  expiration_days          = 7*365
+  retention_policy = {
+    is_locked             = true
+    retention_period_days = 6 * 365
```
xingao267, line 92: what's the different between the `retention_period_days` here and `expiration_days` above? and why are they set to different values?

templates/tfengine/components/audit/main.tf
```diff
@@ -40,84 +40,67 @@ resource "google_{{.parent_type}}_iam_audit_config" "config" {
...
-  ]
+  source  = "terraform-google-modules/log-export/google//modules/storage"
+  version = "~> 5.0.0"
+
+  storage_bucket_name      = "{{.logs_storage_bucket.name}}"
+  project_id               = module.project.project_id
+  location                 = "{{.storage_location}}"
+  log_sink_writer_identity = "${module.storage_export.writer_identity}"
+  storage_class            = "COLDLINE"
+  expiration_days          = 7*365
```
xingao267, line 89: nit: `7 * 365` (with the spaces)

templates/tfengine/components/audit/main.tf
umairidris: dataViewer is done by the project iam permission below (made it bigquery.user)

Converting BQ permissions away from default is being tracked by https://github.com/terraform-google-modules/terraform-google-log-export/issues/49

templates/tfengine/components/audit/main.tf
umairidris: This one's done by the cft module

templates/tfengine/components/audit/main.tf
umairidris: Done.

templates/tfengine/components/audit/main.tf
umairidris: Retention period does not actually do anything. It just forces the bucket to permanently have an expiry that is at least this amount. See https://cloud.google.com/storage/docs/bucket-lock. I also added a comment.

p.s. there seems to be an org policy we should probably set as well: https://cloud.google.com/storage/docs/org-policy-constraints#audit-logging. Do you want to look into that?

xingao267: sure i'll look into that.

xingao267 (approved):

PR #614, opened 2020-10-17
fix(renovate): add fileMatch to regex manager

rarkins:
Adds missing `fileMatch` field to regex manager and removes unnecessary terraform fileMatch.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,11 +9,9 @@
       "groupName": "terraform"
     }
   ],
-  "terraform": {
-    "fileMatch": ["\\.tf$"]
-  },
   "regexManagers": [
     {
+      "fileMatch": ["\\.tf$"],
       "matchStrings": [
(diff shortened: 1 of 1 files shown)


MartinPetkov:
Please test this somehow to make sure it works, I don't want to keep trying endless PRs and have our CI blocked until it does. I maintain that it is always preferable that we first return to a known-good state instead of trying to fix forward.

As an example, for https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/610, it should be changing these files together:
* examples/tfengine/generated/team/example-prod-apps/main.tf
* templates/tfengine/components/resources/gke_clusters/main.tf

rarkins:
It was not known good - it was a misconfiguration causing errors and excessive use on our backend.

MartinPetkov:
It was known good to us - it did what it was supposed to, and the current configuration doesn't. 

I'll merge this PR just to try it, but please in the future test (or suggest way of testing) such PRs to avoid causing a disruption in our repo, as the previous PR has done.

MartinPetkov (approved):

PR #616, opened 2020-10-19
support custom SA in GKE

xingao267:


templates/tfengine/components/resources/gke_clusters/main.tf
```diff
@@ -33,6 +33,9 @@ module "{{resourceName . "name"}}" {
   skip_provisioners          = true
   enable_private_endpoint    = false
   release_channel            = "STABLE"
+  {{if has . "service_account" -}}
```
umairidris, line 36: you can use {{hclField . "service_account"}} here to make this one line

umairidris (approved):

templates/tfengine/components/resources/gke_clusters/main.tf
xingao267: the field in the module is called `compute_engine_service_account`, so I have to explicitly do an if statement here to covert the field name.

umairidris: Ah, right.

PR #619, opened 2020-10-21
Add initial app foundation

umairidris:
Some deployments are copied into components/app for now to make development easier. Eventually if this becomes the defacto way of writing apps we will merge them back with the originals.


examples/tfengine/app.hcl
```diff
@@ -0,0 +1,51 @@
...
+# NOTE: This example is still under development and is not considered stable!
+# Breaking changes to components for this example will not be part of releases.
+
+template "app" {
+  recipe_path = "../../templates/tfengine/recipes/app.hcl"
+  data = {
+    constants = {
+      shared = {
+        env_code        = "s"
+        folder_id       = "123"
```
xingao267, line 24: Should we still support org level deployment?

examples/tfengine/app.hcl
```diff
@@ -0,0 +1,51 @@
...
+
+# NOTE: This example is still under development and is not considered stable!
+# Breaking changes to components for this example will not be part of releases.
+
+template "app" {
+  recipe_path = "../../templates/tfengine/recipes/app.hcl"
+  data = {
+    constants = {
+      shared = {
+        env_code        = "s"
```
MartinPetkov, line 23: Is it described somewhere what this env_code is and what values it can take? I couldn't find a description of it in this PR.

templates/tfengine/components/app/cicd/variables.tf
```diff
@@ -0,0 +1,36 @@
...
+limitations under the License. */ -}}
+
+variable "branch_name" {
+  description = "Branch to trigger on"
+  type        = string
+  default     = "^main$"
+}
+
+variable "source_repo_name" {
+  description = "Name of source repo."
```
MartinPetkov, line 22: Clarify what the format of this is, and maybe give an example. At the moment it's not clear to me if this is a GitHub repo name, or CSR, and if it's a full URL or not, etc.

examples/tfengine/app.hcl
umairidris: I don't think so, apps only need to be deployed in folders.

templates/tfengine/components/app/devops/main.tf
```diff
@@ -0,0 +1,94 @@
...
+    "servicenetworking.googleapis.com",
+    "serviceusage.googleapis.com",
+    "sourcerepo.googleapis.com",
+    "sqladmin.googleapis.com",
+  ]
+}
+
+# Terraform state bucket, hosted in the devops project.
+module "state_bucket" {
+source  = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
```
MartinPetkov, line 75: Whitespace in the front.

templates/tfengine/components/app/cicd/main.tf
```diff
@@ -0,0 +1,194 @@
...
+    repo_name   = google_sourcerepo_repository.configs.name
+    branch_name = var.branch_name
+  }
+  filename = "${local.terraform_root_prefix}cicd/configs/tf-plan.yaml"
+  substitutions = {
+    _TERRAFORM_ROOT = local.terraform_root
+  }
+}
+
+resource "google_cloudbuild_trigger" "apply" {
```
MartinPetkov, line 177: Am I correct in understanding that this trigger will always be created, but will fail every time unless the right permissions are granted to the devops SA to apply changes?

examples/tfengine/app.hcl
umairidris: I will be adding a schema for this section later on but the main thing is that it should be a single letter usually and will be appended as a suffix to globally unique resources (such as projects and buckets).

templates/tfengine/components/app/cicd/main.tf
umairidris: The permissions are granted on line 119

MartinPetkov: I meant that if folder_roles doesn't have permissions to apply, this trigger will fail, right?

templates/tfengine/components/app/cicd/variables.tf
MartinPetkov: You're right that the default provides an example. But I still think the description should have something like your second sentence mentioning that it will create a CSR repo with that name in the project.

templates/tfengine/components/app/cicd/main.tf
umairidris: Oh right. Well the deployment itself will fail

MartinPetkov: Is there a reasonable way to disable this trigger if the permissions aren't there?

umairidris: I can't think of one. If the user does not have permission to grant the SA access on the folder then it should give a pretty clear warning. If there is a permission on the SA missing then when the trigger runs it will highlight the problem there, so I think we are covered in most cases.

MartinPetkov (approved):

xingao267 (approved):

PR #627, opened 2020-10-29
Allow creating projects and networks through app recipe

umairidris:
#625


templates/tfengine/components/resources/compute_networks/main.tf
```diff
@@ -29,7 +29,11 @@ module "{{$resource_name}}" {
     {
```
MartinPetkov, line 29: Judging from the build failure, I think the `{{if has . "subnets" -}}` needs an else clause that sets subnets to an empty list.

umairidris: Yeah, I just set a subnet and made them required, since it doesn't seem like there should be networks without subnets.

examples/tfengine/app.hcl
```diff
@@ -39,6 +39,21 @@ template "app" {
     deployments = [
      {
        name = "project_networks",
+       resources = {
+        project = {
```
xingao267, line 43: indentation

xingao267 (approved):

MartinPetkov (approved):

PR #631, opened 2020-11-03
copy even if formatting or license add fails

### LINKED ISSUE #629: Always generate configs even formatting or validation failed

umairidris:
Fix #629

### DIFF
--- a/internal/tfengine/tfengine.go
+++ b/internal/tfengine/tfengine.go
@@ -95,7 +95,7 @@ func Run(confPath, outPath string, opts *Options) error {
 
 	if opts.Format {
 		if err := hcl.FormatDir(&runner.Default{Quiet: true}, outPath); err != nil {
-			return err
+			return fmt.Errorf("format output dir: %v", err)
 		}
 	}
 
--- a/internal/tfengine/tfengine.go
+++ b/internal/tfengine/tfengine.go
@@ -16,6 +16,7 @@
 package tfengine
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


MartinPetkov:
Failing gen_check.sh

internal/tfengine/tfengine.go
```diff
@@ -79,23 +80,32 @@ func Run(confPath, outPath string, opts *Options) error {
...
 		if err := licenseutil.AddLicense(tmpDir); err != nil {
-			return fmt.Errorf("add license header: %v", err)
+			errs = append(errs, fmt.Sprintf("add license header: %v", err))
 		}
 	}
 
-	if err := os.MkdirAll(outPath, 0755); err != nil {
-		return fmt.Errorf("failed to mkdir %q: %v", outPath, err)
+	if opts.Format {
+		if err := hcl.FormatDir(&runner.Default{Quiet: true}, tmpDir); err != nil {
```
MartinPetkov, line 96: This seems different from the original intent. Wasn't the idea to format after copying?

umairidris: Yes, but the issue is when I do that sometimes it starts formatting unexpected files, since --output_path does not need to contain only the files being generated.

MartinPetkov: Oh, I see. So now it will copy, but surface formatting errors?

umairidris: Right. It will still exit non-zero to show it isn't successful but the files should all still get moved over

MartinPetkov (approved):

xingao267 (approved):

PR #632, opened 2020-11-03
Support more resources in app foundation

umairidris:
#625


xingao267:
> Can't see any obvious issues, but it's a big-ish CL, so I recommend having Xin look as well.

looking

examples/tfengine/app.hcl
```diff
@@ -20,46 +20,267 @@ template "app" {
...
+  length = 16
+  special = true
+}
+EOF
+        }
+      },
       {
         name = "project_networks",
         resources = {
           project = {
```
xingao267, line 77: I like this. Treating the project just like a resource, and we can also make output of this project resource available to other resources.

examples/tfengine/app.hcl
```diff
@@ -20,46 +20,267 @@ template "app" {
...
+          }]
+          cloud_sql_instances = [{
+            name               = "example-mysql-instance"
+            type               = "mysql"
+            network_project_id = "example-prod-networks"
+            network            = "example-network"
+            labels = {
+              type = "no-phi"
+            }
+            # TODO(user): Uncomment and re-run the engine after deploying secrets.
```
xingao267, line 175: There is another pattern which I used in the recent solution, I'll send you the link internally.

examples/tfengine/app.hcl
```diff
@@ -20,46 +20,267 @@ template "app" {
...
+        folder_id           = "123"
+        billing_account     = "000"
+        project_prefix      = "example-prefix"
+        state_bucket        = "example-state"
+        bigquery_location   = "us-east1" # No BigQuery region in us-central1.
+        cloud_sql_region    = "us-central1"
+        cloud_sql_zone      = "a"
+        compute_region      = "us-central1"
+        gke_region          = "us-central1"
+        storage_location    = "us-central1"
```
xingao267, line 33: maybe at somepoint we should standardize the usage of both location and region. Maybe just use `region` and `zone` across all of these variables.

templates/tfengine/recipes/resources.hcl
```diff
@@ -648,7 +690,21 @@ schema = {
             type        = "string"
           }
           network_project_id = {
```
xingao267, line 692: is there way to just define common attributes for different types of resources just once?

umairidris: I can't think of a good way. Maybe we can add reference type of support so we don't need to redefine the fields and can just say this was already defined in an upper layer?

examples/tfengine/app.hcl
umairidris: The reason they are location in some cases is because it can be both regional or multiregional (in the case of storage) whereas for some it can only be region

xingao267: Got it. Maybe we can leave a comment for that.

templates/tfengine/recipes/resources.hcl
xingao267: Yea, maybe not in an upper layer. I was thinking if it is possible to do it in a way that does not affect the end user config, like users will not have another nested block for network_project_id and network_project_suffix and other common stuff, but just reduce attribute definition duplications in the schema. We can leave it as is, I was just wondering if that's even possible.

xingao267 (approved):

templates/tfengine/recipes/resources.hcl
umairidris: Actually I think it is possible. We could define this kinda block at the top level data block, and then just make sure we do something like

network_project := {{get . "network_project_suffix" (get $ "network_project_suffix)}}

PR #635, opened 2020-11-06
support replacing more chars in resourceName func

### LINKED ISSUE #633: Support normalizing dot in resource names

xingao267:
Fixes #633

### DIFF
--- a/examples/tfengine/generated/team/example-prod-data/main.tf
+++ b/examples/tfengine/generated/team/example-prod-data/main.tf
@@ -154,11 +154,11 @@ module "project_iam_members" {
   }
 }
 
-module "example_prod_bucket" {
+module "buckets_example_com" {
   source  = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
   version = "~> 1.4"
 
-  name       = "example.prod.bucket"
+  name       = "buckets.example.com"
   project_id = module.project.project_id
   location   = "us-central1"
 
--- a/examples/tfengine/generated/team/example-prod-data/main.tf
+++ b/examples/tfengine/generated/team/example-prod-data/main.tf
@@ -154,11 +154,11 @@ module "project_iam_members" {
   }
 }
 
-module "buckets_example_com" {
+module "example_prod_bucket" {
   source  = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
   version = "~> 1.4"
 
-  name       = "buckets.example.com"
+  name       = "example-prod-bucket"
   project_id = module.project.project_id
   location   = "us-central1"
 
--- a/examples/tfengine/generated/team/example-prod-data/main.tf
+++ b/examples/tfengine/generated/team/example-prod-data/main.tf
@@ -158,7 +158,7 @@ module "example_prod_bucket" {
   source  = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
   version = "~> 1.4"
 
-  name       = "example-prod-bucket"
+  name       = "example.prod.bucket"
   project_id = module.project.project_id
   location   = "us-central1"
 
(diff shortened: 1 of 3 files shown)


examples/tfengine/generated/team/example-prod-data/main.tf
```diff
@@ -158,7 +158,7 @@ module "example_prod_bucket" {
   source  = "terraform-google-modules/cloud-storage/google//modules/simple_bucket"
   version = "~> 1.4"
 
-  name       = "example-prod-bucket"
+  name       = "example.prod.bucket"
```
MartinPetkov, line 161: Naming buckets with dots may be an issue. From [here](https://cloud.google.com/storage/docs/naming-buckets) and [here](https://cloud.google.com/storage/docs/domain-name-verification):
```
Names containing dots require verification. 

Cloud Storage considers bucket names that contain dots to be domain names. As such, a bucket containing a dot in its name must:

* Be a syntactically valid DNS name (for example, bucket..example.com is not valid because it contains two dots in a row).
* End with a currently-recognized top-level domain, such as .com.
* Be created by a user who is authorized to work with the domain name.

Examples of valid domain-named buckets include example.com, buckets.example.com, and www.example.co.uk
```

xingao267: Thanks for raising this! I changed the example bucket name and added a comment.

internal/template/funcmap.go
```diff
@@ -117,7 +118,13 @@ func resourceName(m map[string]interface{}, key string) (string, error) {
 	if !ok {
 		return "", fmt.Errorf("resource name value %v is not a string", v)
 	}
-	return strings.Replace(name, "-", "_", -1), nil
+
+	r := strings.NewReplacer(
```
umairidris, line 122: optional you could use the dpt v2 way of doing this too:


```
// invalidIDRE defines the invalid characters not allowed in terraform resource names.
var invalidIDRE = regexp.MustCompile("[^a-z0-9-_]")

func resourceName(id string) string {
	return invalidIDRE.ReplaceAllString(strings.ToLower(id), "_")
}
```

umairidris (approved):

examples/tfengine/generated/team/example-prod-data/main.tf
MartinPetkov: I feel like our example bucket names should just not use dots, since there's no need for us to introduce this requirement. It seems like this is only for if you want to use your bucket as a website.

internal/template/funcmap.go
xingao267: Done.

examples/tfengine/generated/team/example-prod-data/main.tf
xingao267: Sure. We can keep the example just having -

MartinPetkov (approved):

internal/template/funcmap.go
```diff
@@ -117,7 +119,10 @@ func resourceName(m map[string]interface{}, key string) (string, error) {
 	if !ok {
 		return "", fmt.Errorf("resource name value %v is not a string", v)
 	}
-	return strings.Replace(name, "-", "_", -1), nil
+
+	// invalidIDRE defines the invalid characters not allowed in terraform resource names.
+	invalidIDRE := regexp.MustCompile("[^a-z0-9_]")
```
umairidris, line 124: this should be defined in the global scope so its only compiled once and erroor is checked then

xingao267: Done.

umairidris: p.s. 'Must' funcs usually imply they will panic on error which is why you usally want to do them in the global scope. If doing inside a func then call the non-Must version and check error manually

xingao267: Got it. Thanks for the info.

PR #639, opened 2020-11-11
Fix Forseti for 0.13

umairidris:
Don't create network in monitor project, instead use the shared vpc host project to create its network. This is how it should be anyway, but since monitoring was done before networking we didn't migrate it over until now.

Now that the network and forseti instance are created in two different deployments it should work. Note that the initial plan will fail until the network project has been created.

### DIFF
--- a/docs/tfengine/schemas/monitor.md
+++ b/docs/tfengine/schemas/monitor.md
@@ -34,12 +34,6 @@ Name of the bastion host's network.
 
 Type: string
 
-### forseti.network_project_id
-
-Name of network project. If unset, will use the current project.
-
-Type: string
-
 ### forseti.security_command_center_source_id
 
 Security Command Center (SCC) Source ID used for Forseti notification.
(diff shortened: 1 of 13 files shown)


docs/tfengine/schemas/monitor.md
```diff
@@ -28,6 +28,12 @@ Domain for the Forseti instance.
 
 Type: string
 
+### forseti.network
+
+Name of the bastion host's network.
```
MartinPetkov, line 33: Copy-paste error, should be "Forseti", not "bastion host". Same in `forseti.subnet`.

templates/tfengine/recipes/monitor.hcl
```diff
@@ -36,6 +29,14 @@ schema = {
           description = "Domain for the Forseti instance."
           type        = "string"
         }
+        network = {
+          description = "Name of the bastion host's network."
```
MartinPetkov, line 33: Copy-past typo, "bastion host"

MartinPetkov (approved):

PR #660, opened 2020-12-01
rm compute service account field on gke cluster

umairidris:
The current config doesn't work. Setting the service account field requires the user to create the SA. By omitting it, the module will create its own SA for the cluster.


xingao267 (approved):
Does the module creates one or use default compute SA?

umairidris:
It creates one

PR #669, opened 2020-12-03
add initial app foundation

umairidris:
This is a proof of concept of the app foundation proposal.

It currently only supports gke clusters.

See https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/tree/foundation/examples/tfengine/generated/app_foundation

### DIFF
--- a/examples/tfengine/app_foundation.hcl
+++ b/examples/tfengine/app_foundation.hcl
@@ -1,0 +1,62 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http:#www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# NOTE: This example is still under development and is not considered stable!
+# Breaking changes to components for this example will not be part of releases.
+
+template "app" {
+  recipe_path = "../../templates/tfengine/recipes/app_foundation.hcl"
+  data = {
+    constants = {
+      shared = {
+        env_code            = "s"
+        folder_id           = "123"
+        billing_account     = "000"
+        project_prefix      = "example-prefix"
+        state_bucket        = "example-state"
+        bigquery_location   = "us-east1" # No BigQuery region in us-central1.
+        cloud_sql_region    = "us-central1"
+        cloud_sql_zone      = "a"
+        compute_region      = "us-central1"
+        gke_region          = "us-central1"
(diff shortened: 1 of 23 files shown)


examples/tfengine/app_foundation.hcl
```diff
@@ -0,0 +1,62 @@
+# Copyright 2020 Google LLC
```
xingao267, line 1: what's the difference between this and existing app.hcl?

umairidris: Existing app.hcl is equivalent to current DPT configs but implements additional best practices around dir structure, constants modules, etc.

This is a POC of the 'patterns' approach, where we would focus on implementing patterns rather than supporting all fields/variations. You can see that there are no projects, apis, etc defined, just resources. Our templates will fill in the rest as needed.

examples/tfengine/app_foundation.hcl
```diff
@@ -0,0 +1,62 @@
...
+        billing_account     = "000"
+        project_prefix      = "example-prefix"
+        state_bucket        = "example-state"
+        bigquery_location   = "us-east1" # No BigQuery region in us-central1.
+        cloud_sql_region    = "us-central1"
+        cloud_sql_zone      = "a"
+        compute_region      = "us-central1"
+        gke_region          = "us-central1"
+        storage_location    = "us-central1"
+        secret_locations    = ["us-central1"]
```
MartinPetkov, line 34: Why is this plural while the others are singular? What does it mean to have multiple secret_locations?

examples/tfengine/app_foundation.hcl
```diff
@@ -0,0 +1,62 @@
...
+        folder_id = "789"
+      }
+    }
+    devops = {
+      admins_group       = "admins@example.com"
+      devops_owners      = ["group:devops-owners-group@example.com"]
+      enable_gcs_backend = false # TODO(user): change to true after initial devops deployment.
+    }
+
+    res = {
```
MartinPetkov, line 52: I find the shortened word to detract from clarity. Is there a reason for this not to be "resources"?

templates/tfengine/recipes/app_foundation.hcl
```diff
@@ -0,0 +1,85 @@
...
+                  }
+                ]
+              },
+              {{end}}
+            ]
+          }]
+        }
+      },
+      {{end}}
+      {{if has .res "gke_clusters"}}
```
MartinPetkov, line 55: This if-statement and the one above can be merged, they're the same.

templates/tfengine/recipes/app_foundation.hcl
```diff
@@ -0,0 +1,85 @@
...
+      {
+        name = "project_networks",
+        resources = {
+          project = {
+            name_suffix        = "networks"
+            is_shared_vpc_host = true
+            apis               = ["compute.googleapis.com"]
+          }
+          compute_networks = [{
+            name = "primary"
```
MartinPetkov, line 32: Is it desirable to have all clusters in the same network by default? Not saying it isn't, but I wanted to get your thoughts.

MartinPetkov (approved):

PR #673, opened 2020-12-07
add groups component

xingao267:


templates/tfengine/components/resources/groups/main.tf
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License. */ -}}
+
+{{range .groups}}
+module "{{resourceName . "id"}}" {
+  source  = "terraform-google-modules/group/google"
+  version = "~> 0.1"
+
+  id = "{{.id}}"
+  customer_id = "{{.customer_id}}"
```
MartinPetkov, line 21: Couldn't these two also just be
```
{{hclField . "id" -}}
{{hclField . "customer_id" -}}
```

MartinPetkov (approved):

templates/tfengine/components/resources/groups/main.tf
xingao267: `hclField` is for optional fields. `id` and `customer_id` are mandatory fields so use {{.id}} and {{.customer_id}} to make sure it is enforced.

examples/tfengine/folder_foundation.hcl
```diff
@@ -88,6 +89,31 @@ template "cicd" {
   }
 }
 
+template "groups" {
+  recipe_path = "{{$recipes}}/project.hcl"
+  output_path = "./groups"
+  data = {
+    project = {
```
umairidris, line 96: Do we need a block like this? Can the groups be included in devops field?

umairidris (approved):

examples/tfengine/folder_foundation.hcl
xingao267: I tried to make it isolated from the rest of the deployment for now so it can be used more generally. In `team.hcl`, I used the group template block to create groups that are used in other deployments (apart from devops, audit, cicd). But I do plan to integrate the creation of groups that are used in devops, audit and cicd in those deployments.

PR #677, opened 2020-12-08
add default gitignore

### LINKED ISSUE #654: Auto-generate a .gitignore file

xingao267:
Fixes #654

### DIFF
--- a/internal/tfengine/tfengine.go
+++ b/internal/tfengine/tfengine.go
@@ -76,6 +76,10 @@ func Run(confPath, outPath string, opts *Options) error {
 	}
 	defer os.RemoveAll(tmpDir)
 
+	if err := addGitIgnore(tmpDir); err != nil {
+		return err
+	}
+
 	if err := dump(c, filepath.Dir(confPath), opts.CacheDir, tmpDir, opts.WantedTemplates); err != nil {
 		return err
 	}
@@ -221,6 +225,20 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
 		if err := write(cp, outputPath, data); err != nil {
 			return fmt.Errorf("component %q: %v", cp, err)
 		}
+	}
+	return nil
+}
+
+func addGitIgnore(path string) error {
+	data := []byte(`# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+`)
+	if err := ioutil.WriteFile(fmt.Sprintf("%s/.gitignore", path), data, 0644); err != nil {
+		return err
 	}
 	return nil
 }
--- a/examples/tfengine/generated/app/.gitignore
+++ b/examples/tfengine/generated/app/.gitignore
@@ -1,0 +1,6 @@
+# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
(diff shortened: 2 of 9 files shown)


internal/tfengine/tfengine.go
```diff
@@ -225,6 +229,20 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
...
 
+func addGitIgnore(path string) error {
+	data := []byte(`# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+`)
+	if err := ioutil.WriteFile(fmt.Sprintf("%s/.gitignore", path), data, 0644); err != nil {
```
MartinPetkov, line 240: You should use `path.Join()` here, not `fmt.Sprintf()`.

MartinPetkov, line 240: This does not need execute permissions, 0444 might make more sense.

internal/tfengine/tfengine.go
```diff
@@ -225,6 +229,20 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
...
 }
 
+func addGitIgnore(path string) error {
+	data := []byte(`# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+`)
```
MartinPetkov, line 239: Have you considered instead using this https://github.com/github/gitignore/blob/master/Terraform.gitignore?

internal/tfengine/tfengine.go
xingao267: Done

xingao267: 0644 doesn't give any execution permissions https://chmodcommand.com/chmod-644/

MartinPetkov: Oh right, my bad.

internal/tfengine/tfengine.go
xingao267: Ignoring .terraform. dirs and state files should cover all temporary files that are generated by DPT operations currently. And we do need terraform.tfvars files to be checked in as they are generated and used by DPT. If users end up modifying their generated Terraform configs, they can add more things in the .gitignore file.

MartinPetkov: Makes sense to me.

MartinPetkov (approved):

internal/tfengine/tfengine.go
```diff
@@ -225,6 +229,20 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
 	return nil
 }
 
+func addGitIgnore(path string) error {
```
umairidris, line 232: Can we make this a component and call it through something like devops recipe?

xingao267: I was initially thinking of doing that but felt that it doesn't really belong to any recipe or component. On a second thought, i think it doesn't make sense to not having any TF configs and still want to ignore tf temporary files. And devops is the most foundational block. I'll add it to the  devops recipe

xingao267: Done

umairidris (approved):

PR #683, opened 2020-12-09
support fetching remote state and use across deployments

### LINKED ISSUE #680: Support create folders and dynamically pass the folder id to project creations

xingao267:
Fixes #680  #683


internal/tfengine/tfengine.go
```diff
@@ -248,6 +255,9 @@ func ConvertToLocalBackend(path string) error {
 		}
 
 		b = backendRE.ReplaceAll(b, nil)
+		b = remoteStateResourceRE.ReplaceAll(b, nil)
+		b = remoteStateDataRE.ReplaceAll(b, []byte("000"))
```
umairidris, line 259: add comment for 000

templates/tfengine/components/folders/outputs.tf
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */ -}}
+
+output "folder_ids" {
+  value = {
+    {{range .folders -}}
+    "{{resourceName . "display_name"}}" = google_folder.{{resourceName . "display_name"}}.folder_id
```
umairidris, line 18: no need for quotes around the keys

templates/tfengine/components/terraform/states/main.tf
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */ -}}
+
+{{range .states -}}
+data "terraform_remote_state" "{{resourceName . "prefix"}}" {
+  backend = "gcs"
+  config = {
+    bucket = "{{$.state_bucket}}"
```
umairidris, line 19: do we want to support the case where state bucket is different?

umairidris (approved):

internal/tfengine/tfengine.go
xingao267: Done

templates/tfengine/components/folders/outputs.tf
xingao267: Done

templates/tfengine/components/terraform/states/main.tf
xingao267: Maybe when there is use case for that? At least for now, when using DPT, the state bucket is set at the top level config and fully managed by DPT.

PR #724, opened 2021-01-06
do not create additional default SA in project

xingao267:


xingao267:
pending https://github.com/terraform-google-modules/terraform-google-project-factory/pull/529 to be merged first

xingao267:
@yeweidaniel FYI

examples/tfengine/generated/org_foundation/devops/main.tf
```diff
@@ -33,13 +33,16 @@ terraform {
...
-  name                    = "example-devops"
-  org_id                  = "12345678"
-  billing_account         = "000-000-000"
-  lien                    = true
+  name            = "example-devops"
+  org_id          = "12345678"
+  billing_account = "000-000-000"
+  lien            = true
+  # Create and keep default service accounts when certain APIs are enabled.
   default_service_account = "keep"
```
umairidris, line 43: how does this differ to default_service_account? Do we need both?

xingao267: We do need both. `default_service_account` controls the behaviour for the real default SAs, such as the default GCE SA, GAE SA, etc.

`create_project_sa` controls whether or not to create an additional service account called "project-service-account".  It is just an additional thing that the Project Factory module does after creating the project. But in reality, we don't use this SA. This new flag was recently introduced to address https://github.com/terraform-google-modules/terraform-google-project-factory/issues/526.

umairidris: Ah, got it. Thanks!

templates/tfengine/components/project/main.tf
```diff
@@ -30,17 +30,22 @@ module "project" {
...
+  folder_id       = "{{.parent_id}}"
   {{- end}}
-  billing_account         = "{{.billing_account}}"
+  billing_account = "{{.billing_account}}"
   {{- end}}
-  lien                    = {{get . "enable_lien" true}}
+  lien            = {{get . "enable_lien" true}}
+  # Create and keep default service accounts when certain APIs are enabled.
   default_service_account = "keep"
+  {{- if not (get . "use_constants" false)}}
```
umairidris, line 45: Is this doing anything? Seems to only have a comment inside?

xingao267: Yea, i need this because if `use_contants` is `true`, there is no `project_id` input as the project ID is constructed from prefix and env code.

xingao267: Just want to dynamically construct this comment to include the project ID as well.

examples/tfengine/generated/org_foundation/devops/main.tf
xingao267: No problem!

templates/tfengine/components/project/main.tf
umairidris: Ah I see. I don't think it is that important to include the full SA name since it can be assumed it will be created in this project. What about changing the wording to be more generic e.g.

Do not create the additional project service account to run Terraform as it will be handled by separate CICD process.

xingao267: Done.

Rephrased a little as the SA is to be used for Compute Engine generically. https://github.com/terraform-google-modules/terraform-google-project-factory#features

umairidris (approved):

PR #725, opened 2021-01-07
Use input dir for temp plan file in import

MartinPetkov:
Terraform may be sandboxed or running on a system that can't write to /tmp.

Also print not importable files in dry-run mode.

### DIFF
--- a/internal/tfimport/tfimport.go
+++ b/internal/tfimport/tfimport.go
@@ -578,7 +578,7 @@ func planAndImport(rn, importRn runner.Runner, runArgs *RunArgs) (retry bool, er
 	var errs []string
 	var skipped []string
 	var importCmds []string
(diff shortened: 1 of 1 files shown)


internal/tfimport/tfimport.go
```diff
@@ -577,6 +578,7 @@ func planAndImport(rn, importRn runner.Runner, runArgs *RunArgs) (retry bool, er
 	var errs []string
 	var skipped []string
 	var importCmds []string
+	var notImportable []string
```
xingao267, line 581: maybe name it as `notImportableMsgs`?

MartinPetkov: Done.

xingao267 (approved):

PR #727, opened 2021-01-07
add caller as default owner in devops groups if none is specified

### LINKED ISSUE #715: Support adding caller as an initial owner to groups

xingao267:
Fixes #715

### DIFF
--- a/templates/tfengine/components/devops/main.tf
+++ b/templates/tfengine/components/devops/main.tf
@@ -97,7 +97,7 @@ module "owners_group" {
   {{- end}}
   {{hclField .project.owners_group "description" -}}
   {{if has .project.owners_group "owners" -}}
-  owners = {{hcl .project.owners_group.owners}}
+  owners = {{hcl .project.owners_group.owners -}}
   {{else -}}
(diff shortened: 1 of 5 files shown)


MartinPetkov (approved):

templates/tfengine/components/devops/main.tf
```diff
@@ -125,7 +131,11 @@ module "admins_group" {
   display_name = "{{regexReplaceAll "@.*" .admins_group.id ""}}"
   {{- end}}
   {{hclField .admins_group "description" -}}
-  {{hclField .admins_group "owners" -}}
+  {{if has .admins_group "owners" -}}
```
umairidris, line 134: Can we simply not create the owners group and just leave the default owner on the project (which will be the caller)?

xingao267: I think it's better to always create a placeholder group during initial devops deployment and assign owner role to it so future owner permission granting does not necessarily go though IAM but through adding to the group.

And for org/folder admin role, we have to explicitly grant the IAM role to the group/individual anyway. So it's more consistent to do the same for devops owner role.

umairidris (approved):

PR #731, opened 2021-01-08
update upstream policy library templates

xingao267:
Due to https://github.com/forseti-security/policy-library/pull/390


MartinPetkov (commented):
How come this PR only affects the generated files? Does this come from upstream during generation?

xingao267:
Yes. It is from this step [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/scripts/regen.sh#L26). `policygen` fetches upstream policy library templates at head and save it in the example folder.

MartinPetkov (approved):

PR #738, opened 2021-01-13
make istio in gke configurable

xingao267:
It's nice to have for microservices. Does not need to be on by default
in all cases. https://cloud.google.com/istio/docs/istio-on-gke/overview

Also see [Should I use Istio on GKE?](https://cloud.google.com/istio/docs/istio-on-gke/overview#should_i_use). "Due to these limitations and the product's beta status, customers should not use this add-on in production. Anthos Service Mesh or open source Istio are better options for production workloads."

### DIFF
--- a/docs/tfengine/schemas/resources.md
+++ b/docs/tfengine/schemas/resources.md
@@ -629,6 +629,12 @@ Name of the secondary subnet range to use for services.
 
 Type: string
 
+### gke_clusters.istio
+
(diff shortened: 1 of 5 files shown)


MartinPetkov (approved):

MartinPetkov:
The "Don't use it in production, use ASM or OSS Istio" recommendation is new to me. I'll give that part of the article a read.

Issue #658, opened 2020-11-30
Extend CICD to support GKE app deployment


xingao267:

xingao267:
Verify kubernetes_* resources can be successfully deployed and fix any issues occurred.

xingao267:
Container service agent needs permission in the host network project to create additional firewall rules for health check. Upstream PR sent for review https://github.com/terraform-google-modules/terraform-google-project-factory/pull/536

umairidris:
I thought the regular project factory module can now also create vpc attachments, so why do we still need to use the shared vpc module? It seems to always get out of sync.

xingao267:
@umairidris I think that's separate issue tracked in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/697. Neither the PF main module nor the service project module supports that field at this moment.

umairidris:
Ah right, thanks for clarifying. We should look into fixing that for the next major release (might cause some existing projects to become recreated tho).

xingao267:
Another issue in the module: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/789

PR #742, opened 2021-01-14
enable grant_services_security_admin_role feature

### LINKED ISSUE #740: GKE app failed to start due to Load Balancer health check firewall rule cannot be created in VPC host project
This issue has happened few times. Just want to document the issue and solutions here. 

When using DPT or CFT's GKE modules to create the cluster, it also adds health check. And when the VPC is hosted in a separate project, the GKE Service Agent fail to create a firewall rule for HC and potentially other ones (not always predictable). The support has been added in PF module as https://github.com/terraform-google-modules/terraform-google-project-factory/pull/483 and https://github.com/terraform-google-modules/terraform-google-project-factory/pull/536. We need to include that in DPT.

xingao267:
Fixes #740
#658

### DIFF
--- a/examples/tfengine/generated/app/modules/project_apps/main.tf
+++ b/examples/tfengine/generated/app/modules/project_apps/main.tf
@@ -38,6 +38,9 @@ module "project" {
   default_service_account = "keep"
   # Do not create an additional project service account to be used for Compute Engine.
   create_project_sa = false
+  # When Kubernetes Engine API is enabled, grant Kubernetes Engine Service Agent the
+  # Compute Security Admin role on the VPC host project so it can manage firewall rules.
+  grant_services_security_admin_role = true
 
   svpc_host_project_id = "${local.constants.project_prefix}-${local.constants.env_code}-networks"
   activate_apis        = ["compute.googleapis.com"]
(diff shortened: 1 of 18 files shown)


umairidris (commented):
I think we should make this a manually configurable field. It only applies to the apps project. You could also force set it to true when gke clusters are defined.

xingao267:
The underlying Project Factory module already imposes a [condition](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/shared_vpc_access/main.tf#L89) on whether or not container.googleapis.com is enabled. I don't think it's necessary for us to do a check again here. We can do the same type of check based on whether the API is enabled, but that's duplicated of what has been done by PF.

xingao267:
Also given project creation and resources creation can be decoupled in separate deployments, and customers might use `raw_configs` field to use other type of GKE cluster module. It's difficult to know whether or not a cluster is requested based on resources defined in DPT.

umairidris:
Ok, the naming of these variables is confusing. IIUC, what `grant_services_security_admin_role` do is simply *allow* the PF module to grant the role IF the correct API is enabled? If there is no API then it is a no-op?

xingao267:
> Ok, the naming of these variables is confusing. IIUC, what `grant_services_security_admin_role` do is simply _allow_ the PF module to grant the role IF the correct API is enabled? If there is no API then it is a no-op?

Yes. Only when correct API (currently, only the GKE API) is enabled. Noop when it's not enabled.

umairidris (approved):

umairidris:
Maybe clarify it is a no-op as a comment.

xingao267:
> Maybe clarify when it is a no-op as a comment.

Done

PR #748, opened 2021-01-21
Add Multi runner for tfimport to show better logs

MartinPetkov:
For tfimport, it is useful to print the terraform commands as they are happening, as they may freeze but not show any logs. This can happen if they're waiting for input and maybe if the gcloud authentication cookie has expired.


internal/runner/runner.go
```diff
@@ -72,6 +73,59 @@ func (d *Default) CmdCombinedOutput(cmd *exec.Cmd) ([]byte, error) {
...
+
+// CmdRun executes the command aand prints stdout and stderr without returning either.
+func (d *Multi) CmdRun(cmd *exec.Cmd) error {
+	if !d.Quiet {
+		log.Printf("Running: %v", cmd.Args)
+	}
+
+	var stderr bytes.Buffer
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = io.MultiWriter(os.Stderr, &stderr)
```
umairidris, line 91: why do we need this? doesn't the output stream to stderr anyway so it is visible to the user? this will cause the stderr to be printed twice (once through stderr then again through the error msg)

internal/runner/runner.go
```diff
@@ -72,6 +73,59 @@ func (d *Default) CmdCombinedOutput(cmd *exec.Cmd) ([]byte, error) {
...
+		log.Printf("Running: %v", cmd.Args)
+	}
+
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = io.MultiWriter(os.Stdout, &stdout)
+	cmd.Stderr = io.MultiWriter(os.Stderr, &stderr)
+
+	err := cmd.Run()
+	if err != nil {
+		err = fmt.Errorf("%v: %s", err, stderr.String())
```
umairidris, line 111: ditto, if stderr is visible to user no need to append it here

umairidris (approved):

internal/runner/runner.go
MartinPetkov: You're right. I'll change it.

PR #756, opened 2021-01-22
Fix APIs not being enabled when exists = true

### LINKED ISSUE #755: API field enables list of APIs only when project_exist = false
When project_exist = true, then API enablement fails.

Example:
project = {​​​​​​​
      project_id = "aw-8f00ef2f-2e8c-49e5-9635-46a"
      exists     = true
      api = [
         
      ]
     
    }​​​​​​​

We have an existing project in which we deploy resources, hence we need to be able to enable API for an existing project.

umairidris:
Fix #755 

- Add module "project" with source pointing to project-services when exists is true. This allows us to create a dependency on it when creating resources.
- Remove unused project outputs

Note: the "groups" deployments look a bit weird now because they use the project.hcl recipe so they will usually have an empty APIs module. We could rework it in the future to make it look cleaner.


MartinPetkov (approved):

templates/tfengine/components/resources/bastion_hosts/main.tf
```diff
@@ -18,7 +18,7 @@ module "{{resourceName . "name"}}" {
   version = "~> 3.0.0"
 
   name         = "{{.name}}"
-  project      = {{- if get $.project "exists" false}} "{{$.project.project_id}}" {{- else}} module.project.project_id {{end}}
+  project      =  module.project.project_id
```
xingao267, line 21: extra space

xingao267 (approved):

PR #773, opened 2021-02-04
add a sleep between group creation and IAM bindings

xingao267:
The group is not ready for IAM bindings right after the
group resource is deployed.

### DIFF
--- a/templates/tfengine/components/devops/main.tf
+++ b/templates/tfengine/components/devops/main.tf
@@ -104,13 +104,25 @@ module "owners_group" {
     module.project
   ]
 }
+
+resource "time_sleep" "owners_wait" {
+  depends_on = [
+    module.owners_group,
+  ]
+  create_duration = "10s"
+}
 {{- end}}
 
 # Project level IAM permissions for devops project owners.
 resource "google_project_iam_binding" "devops_owners" {
   project = module.project.project_id
   role    = "roles/owner"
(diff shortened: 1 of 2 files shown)


templates/tfengine/components/devops/main.tf
```diff
@@ -104,13 +104,25 @@ module "owners_group" {
     module.project
   ]
 }
+
+resource "time_sleep" "owners_wait" {
```
MartinPetkov, line 108: You should add a comment above this resource explaining why it's needed.

templates/tfengine/components/devops/main.tf
```diff
@@ -104,13 +104,25 @@ module "owners_group" {
...
+  depends_on = [
+    module.owners_group,
+  ]
+  create_duration = "10s"
```
MartinPetkov, line 112: Is 10s definitely more than enough? It should be enough that this is essentially never flaky.

MartinPetkov (approved):

templates/tfengine/components/devops/main.tf
xingao267: I tried 5s before and was a bit flaky so changed to 10s. Let me do 15s to be safe.

templates/tfengine/components/devops/main.tf
xingao267: Done

PR #777, opened 2021-02-04
support importing projects in automation

xingao267:


examples/tfengine/generated/org_foundation/cicd/configs/import.sh
```diff
@@ -0,0 +1,45 @@
...
+IFS=' ' read -r -a DIRS <<< "${DIRS}"
+
+for mod in "${DIRS[@]}"
+do
+    cd "${ROOT}"/"${mod}"
+    terraform init
+    terraform plan -out=plan.tfplan
+    project_id=$(terraform show -json plan.tfplan | jq -rM '.resource_changes[]? | select(.change.actions | index("create")) | select(.address | index("module.project.module.project-factory.google_project.main")) | .change.after.project_id')
+    if ! [[ -z "${project_id}" ]]; then
+      terraform import module.project.module.project-factory.google_project.main ${project_id}
```
umairidris, line 43: Can we use tfimport? Seems that would scale better.

xingao267: I was initially trying that. But it seems the --interactive=false mode does not working fully as expected (was able to skip something but not all), so turned back to a single import command for now. Also since it's going to be run as part of the automation, I would like to be able to control what type of resources I would like to import instead of going through the whole thing which could increase the chance of errors. I can open some issues for tfimport to fix the interactive issue as well support a type list argument and then use it here.

umairidris: Would it be useful to add a --resources flag that would allow a user to specify a list of resources to look at?

So then you could do tfimport --resources=google_project --nointeractive

xingao267: @MartinPetkov another data point for you. I tried both --nointeractive and --interactive=false, but it seems to stop at the lien resource

```
$ tfimport . --nointeractive
2021/02/04 18:01:47 Running: [terraform init]
2021/02/04 18:01:52 Running: [terraform plan -out plan-for-import-795646437.tfplan]
2021/02/04 18:01:56 Running: [terraform show -json plan-for-import-795646437.tfplan]
2021/02/04 18:01:58 Found importable resource: "module.project.module.project-factory.google_resource_manager_lien.lien[0]"
2021/02/04 18:01:59 Could not determine "name" automatically
2021/02/04 18:01:59 No existing liens found in 488909503971
Enter a value manually (blank to skip):

```

But if we can support a `--resources` array then, we can work around this problem as well.

MartinPetkov: If --interactive=false doesn't work, that's definitely a bug. I'll fix that first.

Having a --resources flag should be possible. Let me look into it after.

MartinPetkov: Done:
* https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/778
* https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/779

xingao267: Done

examples/tfengine/generated/org_foundation/cicd/configs/import.sh
```diff
@@ -0,0 +1,47 @@
...
+
+VERSION=v0.6.0
+wget -q -O ${ROOT}/tfimport https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/releases/download/${VERSION}/tfimport_${VERSION}_linux-amd64
+chmod +x ${ROOT}/tfimport
+
+for mod in "${DIRS[@]}"
+do
+    cd "${ROOT}"/"${mod}"
+    ${ROOT}/tfimport -input_dir "${ROOT}"/"${mod}" --resource_types 'google_project' --resource_types 'google_project_service' --interactive false || true
+    rm -rf .terraform*
```
umairidris, line 44: No need for rm

umairidris (approved):

examples/tfengine/generated/org_foundation/cicd/configs/import.sh
xingao267: Done

PR #778, opened 2021-02-05
Support interactive correctly

MartinPetkov:


internal/tfimport/importer/google_resource_manager_lien.go
```diff
@@ -30,6 +30,11 @@ type ResourceManagerLien struct{}
 
 // ImportID returns the ID of the resource to use in importing.
 func (i *ResourceManagerLien) ImportID(rc terraform.ResourceChange, pcv ConfigMap, interactive bool) (string, error) {
+	// Can't import if not interactive, since lien names are system-generated and will never be in the plan.
+	if !interactive {
```
umairidris, line 34: optional: it seems like a lot of resources will match this criteria so we will need to keep repeating this logic for it. Could we just add a bool field to indicate this can't be imported automatically and then do a check for it once in the base tfimport code?

MartinPetkov: Done, added a common code path for this. However, in the interest of defense-in-depth, since the ImportID function takes the interactive flag, I'd like to leave those "just in case" checks in there too.

umairidris: IMO less code == fewer chances for bugs ;) 

but this is fine

Does it still need the interactive flag if common path handles it?

internal/tfimport/tfimport.go
```diff
@@ -466,12 +472,17 @@ func (ir Resource) ImportID(interactive bool) (string, error) {
...
 // pcv represents provider config values, which will be used if the resource does not have values defined.
-func Importable(rc terraform.ResourceChange, pcv importer.ConfigMap) (*Resource, bool) {
+func Importable(rc terraform.ResourceChange, pcv importer.ConfigMap, interactive bool) (*Resource, bool) {
 	ri, ok := Importers[rc.Kind]
 	if !ok {
 		return nil, false
 	}
 
+	if _, reqInteractive := RequiresInteractive[rc.Kind]; reqInteractive && !interactive {
+		fmt.Printf("Martin found %v requires interactive\n", rc.Kind)
```
umairidris, line 482: Martin found ? :)

umairidris (approved):

internal/tfimport/tfimport.go
MartinPetkov: Ah yes. I purposely put my name in to make it glaringly obvious when I forget to remove log messages :)

PR #779, opened 2021-02-05
Importer: Add resource_types flag

MartinPetkov:


cmd/tfimport/main.go
```diff
@@ -16,7 +16,7 @@
 // Requires Terraform to be installed and for authentication to be configured for each provider in the Terraform configs provider blocks.
 //
 // Usage:
-// $ go run . [--input_dir=/path/to/config]
+// $ go run . [--input_dir=/path/to/config] [--resources 'google_storage_bucket.bucket' --resources 'google_resource_manager_lien.lien']
```
umairidris, line 19: I was actually thinking of limiting just based on the resource type e.g. google_storage_bucket. How does one use this when the resource is in a module? Do they need to give a whole module path?

I think if people have to figure out the resource name/full path then it makes the use of tfimport not as useful, since at that point you could just run the `terraform import` command yourself. I would prefer the resource type limit, that should solve the main use cases I imagine...

MartinPetkov: You're right, I've changed it to resource type.

cmd/tfimport/main.go
```diff
@@ -16,7 +16,7 @@
 // Requires Terraform to be installed and for authentication to be configured for each provider in the Terraform configs provider blocks.
 //
 // Usage:
-// $ go run . [--input_dir=/path/to/config]
+// $ go run . [--input_dir=/path/to/config] [--resource_types 'google_storage_bucket.bucket' --resources 'google_resource_manager_lien.lien']
```
umairidris, line 19: Have you tested that listing the resource multiple times works as intended, or do we need to list it once and comma separate?

umairidris (approved):

cmd/tfimport/main.go
MartinPetkov: I have tested, it does work as intended when specified multiple times.

cmd/tfimport/main.go
```diff
@@ -29,6 +29,20 @@ import (
 	"github.com/GoogleCloudPlatform/healthcare-data-protection-suite/internal/tfimport"
 )
 
+// From https://stackoverflow.com/a/28323276
+type mapFlag map[string]bool
+
+func (i *mapFlag) String() string {
+	return "my string representation"
```
umairidris, line 36: Is this supposed to be a repr of the map?

Maybe use json.Marshal?

MartinPetkov: Yes, I forgot to change this line. Will fix it.

cmd/tfimport/main.go
```diff
@@ -16,10 +16,11 @@
 // Requires Terraform to be installed and for authentication to be configured for each provider in the Terraform configs provider blocks.
 //
 // Usage:
-// $ go run . [--input_dir=/path/to/config]
+// $ go run . [--input_dir=/path/to/config] [--resource_types 'google_storage_bucket.bucket' --resources 'google_resource_manager_lien.lien']
```
xingao267, line 19: no need to include the resource name in the argument, and consider updating the README.md as well

MartinPetkov: Opened https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/783 to fix this.

PR #783, opened 2021-02-06
Fix tfimport docs

MartinPetkov:


docs/tfimport/README.md
```diff
@@ -57,14 +57,22 @@ differences between the configs and the actual infrastructure.
...
+        without running them.
   -input_dir string
-     Path to the directory containing Terraform configs. (default ".")
+        Path to the directory containing Terraform configs. (default ".")
   -interactive
-     Interactively ask for user input when import information cannot be
-     automatically determined. (default true)
+        Interactively ask for user input when import information cannot be
+        automatically determined. (default true)
+  -resource_types value
```
xingao267, line 67: change the type to string or list of string?

docs/tfimport/README.md
```diff
@@ -57,14 +57,22 @@ differences between the configs and the actual infrastructure.
...
+        Specific resource types to import, specified as terraform
+        resource names (e.g. 'google_storage_bucket',
+        'google_resource_manager_lien'). Leave empty to import all.
```
xingao267, line 70: also mention that to list multiple resource types, specify this flag multiple times

cmd/tfimport/main.go
```diff
@@ -16,7 +16,7 @@
 // Requires Terraform to be installed and for authentication to be configured for each provider in the Terraform configs provider blocks.
 //
 // Usage:
-// $ go run . [--input_dir=/path/to/config] [--resource_types 'google_storage_bucket.bucket' --resources 'google_resource_manager_lien.lien']
+// $ go run . [--input_dir=/path/to/config] [--resource_types 'google_storage_bucket' --resource_types 'google_resource_manager_lien']
```
xingao267, line 19: remove the `=` between `--input_dir` and `/path/to/config` to be consistent with how `--resource_types` is used. (use space instead of `=`)

docs/tfimport/README.md
xingao267: it seems --dry_run and --interactive did not have type specified, we can also add those

xingao267, line 67: If the examples are using `--` (for the flags), then also use `--` here as well instead of `-`

MartinPetkov: Sure, but normally this is just copy-pasted from the run of `tfimport -h`.

xingao267: hmm, if that's the case, then we probably should use the output from `tfimport -h`

xingao267 (approved):

docs/tfimport/README.md
MartinPetkov: Done.

PR #786, opened 2021-02-08
Build binaries statically

MartinPetkov:
This lets them run in the Alpine-based containers used in the CI/CD templates.

### DIFF
--- a/build/build-include.sh
+++ b/build/build-include.sh
@@ -54,7 +54,7 @@ function build_binaries() {
         bin_name="$(basename ${build_dir})"
         bin="${bin_name}_${VERSION}_${OS}-${ARCH}"
         echo "Building ${OUTPUT_DIR}/${bin}"
-        env GOOS="${OS}" GOARCH="${ARCH}" go build -ldflags="-X 'github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd.Version=${VERSION}' -X 'github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd.BuildTime=${BUILD_TIME}'" -o "${OUTPUT_DIR}/${bin}" "${build_dir}"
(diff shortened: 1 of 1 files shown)


build/build-include.sh
```diff
@@ -54,7 +54,7 @@ function build_binaries() {
         bin_name="$(basename ${build_dir})"
         bin="${bin_name}_${VERSION}_${OS}-${ARCH}"
         echo "Building ${OUTPUT_DIR}/${bin}"
-        env GOOS="${OS}" GOARCH="${ARCH}" go build -ldflags="-X 'github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd.Version=${VERSION}' -X 'github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd.BuildTime=${BUILD_TIME}'" -o "${OUTPUT_DIR}/${bin}" "${build_dir}"
+        env CGO_ENABLED=0 GOOS="${OS}" GOARCH="${ARCH}" go build -ldflags="-extldflags '-static' -X 'github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd.Version=${VERSION}' -X 'github.com/GoogleCloudPlatform/healthcare-data-protection-suite/cmd.BuildTime=${BUILD_TIME}'" -o "${OUTPUT_DIR}/${bin}" "${build_dir}"
```
xingao267, line 57: what does CGO_ENABLED do?

xingao267 (approved):
Let's do a binary v0.7.0 release after this and https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/782 is merged.

build/build-include.sh
MartinPetkov: It's covered here: https://golang.org/cmd/cgo/#hdr-Using_cgo_with_the_go_command

"cgo enables the creation of Go packages that call C code. ... The cgo tool is enabled by default ... and disabled by default when cross-compiling. You can control this by setting the CGO_ENABLED environment variable when running the go tool: set it to 1 to enable the use of cgo, and to 0 to disable it."

We don't use cgo directly, but I went to discover what does:
```
$ go list -f "{{if .CgoFiles}}{{.ImportPath}}{{end}}" ./... $(go list -f "{{.ImportPath}}{{range .Deps}} {{.}}{{end}}" ./...)
crypto/internal/boring
net
os/user
runtime/cgo
```

So now I'm a little concerned. It means if we are accidentally using a function that requires cgo, it won't work. But if we compile with CGO_ENABLED, it doesn't work in CICD.

Are you able to try out the import.sh with a binary built with CGO_ENABLED=0 before I submit this?

xingao267: works!

Issue #781, opened 2021-02-06
tfimport retries manually skipped values


umairidris:
I had a lien and a random string that needed manual intervention. I skipped the random string and imported the lien. Then it retried for some reason. In the case I skip a resource I expect it won't keep retrying it.

```
2021/02/05 20:58:17 Found importable resource: "module.project.module.project-factory.random_id.random_project_id_suffix"
2021/02/05 20:58:17 Could not determine "b64_url" automatically
2021/02/05 20:58:17 Please enter the previously-generated random_id, in *hex* form. See https://www.terraform.io/docs/providers/random/r/id.html#attributes-reference.
Enter a value manually (blank to skip):

2021/02/05 20:58:56 Some imports succeeded but others did not. Retrying the import, in case dependent values have now been populated.
2021/02/05 20:58:56 Running: [terraform init]
2021/02/05 20:59:02 Running: [terraform plan -out deploy_generated/dev/networks/plan-for-import-622910997.tfplan]
2021/02/05 20:59:11 Running: [terraform show -json deploy_generated/dev/networks/plan-for-import-622910997.tfplan]
2021/02/05 20:59:14 Found importable resource: "module.project.module.project-factory.google_resource_manager_lien.lien[0]"
2021/02/05 20:59:14 Could not determine "name" automatically
2021/02/05 20:59:14 Found the following liens in projects/652218318533:
Name                                                 Origin              Reason                                                                                                                                 Restrictions
p652218318533-l70a2fb7d-f26c-4ba7-bfd6-b76f4d2fcbfb  project-factory     Project Factory lien                                                                                                                   resourcemanager.projects.delete
p652218318533-l99a29bf1-c508-4d12-a420-e4e0e648597c  xpn.googleapis.com  This lien is added to prevent the deletion of this shared VPC host project. The host project should be disabled before it is deleted.  resourcemanager.projects.delete
Enter a value manually (blank to skip):
p652218318533-l70a2fb7d-f26c-4ba7-bfd6-b76f4d2fcbfb
2021/02/05 20:59:37 Running: [terraform import module.project.module.project-factory.google_resource_manager_lien.lien[0] 652218318533/p652218318533-l70a2fb7d-f26c-4ba7-bfd6-b76f4d2fcbfb]
[...]

PR #789, opened 2021-02-09
Track skipped resources and re-skip on retry

MartinPetkov:
Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/781

### DIFF
--- a/internal/tfimport/tfimport.go
+++ b/internal/tfimport/tfimport.go
@@ -529,8 +529,6 @@ type RunArgs struct {
 	SpecificResourceTypes map[string]bool
 }
 
-var skipped = make(map[string]bool)
-
 // Run executes the main tfimport logic.
 func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
 	// Expand the config path (ex. expand ~).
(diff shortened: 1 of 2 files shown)


internal/tfimport/tfimport.go
```diff
@@ -529,6 +529,8 @@ type RunArgs struct {
 	SpecificResourceTypes map[string]bool
 }
 
+var skipped = make(map[string]bool)
```
umairidris, line 532: can this be moved inside Run?

MartinPetkov: Yes.

internal/tfimport/tfimport.go
```diff
@@ -550,14 +551,16 @@ func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
 
 		log.Println("Some imports succeeded but others did not. Retrying the import, in case dependent values have now been populated.")
 	}
+	fmt.Printf("Martin skipped: %v\n", skipped)
```
umairidris, line 554: rm

umairidris (approved):

PR #792, opened 2021-02-09
add login instructions

xingao267:


docs/tfengine/README.md
```diff
@@ -257,6 +257,11 @@ directly use the `terraform` binary to deploy the infrastructure.
     that would be checked into source control, i.e. `terraform` in the above
     setup.
 
+1. Login to your Google account by following instructions
+    [here](https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login).
+    Make sure to login using `gcloud auth application-default login` instead of
+    `gcloud auth login`. Terraform uses Application Default Credentials.
```
MartinPetkov, line 263: Don't even mention "instead of ...", in case someone is quickly skimming they might copy the wrong thing. Just say "Make sure to login using `<cmd>` since Terraform uses..."

umairidris (approved):

docs/tfengine/README.md
xingao267: Done

PR #793, opened 2021-02-09
Change error messages based on verbosity and show errs even with skipped resources

MartinPetkov:
Take into account verbosity when printing error summary at the end. Also, show errors when some resources were skipped, this was previously wrong.

### DIFF
--- a/cmd/tfimport/main.go
+++ b/cmd/tfimport/main.go
@@ -93,6 +93,7 @@ func run() error {
 		DryRun:                *dryRun,
 		Interactive:           *interactive,
 		SpecificResourceTypes: resourcesFlag,
+		Verbose:               *verbose,
 	}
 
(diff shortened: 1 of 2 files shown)


umairidris (approved):

xingao267:
> Take into account verbosity when printing error summary at the end. Also, show errors when some resources were skipped, this was previously wrong.

What errors are we showing here when some resources were skipped? Are those the errors from importing other resources?

MartinPetkov:
> > Take into account verbosity when printing error summary at the end. Also, show errors when some resources were skipped, this was previously wrong.
> 
> What errors are we showing here when some resources were skipped? Are those the errors from importing other resources?

Yes, just the normal "failed to import" error messages plus the output from the terraform import command. Previously the logic was "if something was skipped, print skipped resources and return", but that return happened before printing errors.

xingao267 (approved):

PR #796, opened 2021-02-10
Importer: Refactor and add final summary of results

MartinPetkov:
Keep track of succeeded, skipped, and errored imports in the main loop, and report on them at the end of it.

### DIFF
--- a/internal/tfimport/tfimport.go
+++ b/internal/tfimport/tfimport.go
@@ -518,35 +518,35 @@ func DoesNotExist(output string) bool {
 	return reDoesNotExist.FindStringIndex(output) != nil
 }
 
-// importErr is an error indicating a resource failed to be imported.
+// importError is an error indicating a resource failed to be imported.
 // This doesn't indicate an error in running the steps around trying an import,
 // it's specifically for terraform failing to import a resource for a reason
 // that can't be handled (i.e. does not exist or not importable at all).
-type importErr struct {
+type importError struct {
 	// errMgs contains the error messages for all failed imports.
 	errMsgs []string
 }
 
-func (e *importErr) Error() string {
+func (e *importError) Error() string {
 	return fmt.Sprintf("failed to find or import %v resources. Use --verbose flag to see detailed error messages.", len(e.errMsgs))
 }
 
 var summaryTmpl string = `
 Summary:
 
-Found {{.total}} importable resources.
+Found {{.total}} importable resources
 
-Successfully imported {{len .successes}}{{if gt (len .successes) 0}}:{{else}}.{{end}}
+Successfully imported {{len .successes}}{{if gt (len .successes) 0}}:{{end}}
 {{- range $resource := .successes}}
 - {{$resource}}
 {{- end}}
 
-Skipped {{len .skipped}}{{if gt (len .skipped) 0}}:{{else}}.{{end}}
+Skipped {{len .skipped}}{{if gt (len .skipped) 0}}:{{end}}
 {{- range $resource := .skipped}}
 - {{$resource}}
 {{- end}}
 
-Failed to import {{len .failures}}{{if gt (len .failures) 0}}:{{else}}.{{end}}
+Failed to import {{len .failures}}{{if gt (len .failures) 0}}:{{end}}
 {{- range $resource := .failures}}
 - {{$resource}}
 {{- end}}
@@ -575,41 +575,40 @@ func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
 
 	var successesTotal, failuresTotal []string
 	skipped := make(map[string]bool)
-	var ie *importErr
+	var ie *importError
 	for {
 		failuresTotal = nil
 		successes, err := planAndImport(rn, importRn, runArgs, skipped)
 		successesTotal = append(successesTotal, successes...)
 
-		if err != nil {
-			if errors.As(err, &ie) {
-				failuresTotal = ie.errMsgs
-				if len(successes) > 0 {
-					// Some resources imported successfully, but others didn't.
-					// Time to retry.
-					log.Println("Some imports succeeded but others did not. Retrying the import, in case dependent values have now been populated.")
-					continue
-				}
-			} else {
-				// Otherwise it's some other kind of error and we should return it.
-				return err
-			}
+		if err == nil {
+			// Either no successes, or no errors.
+			// No need to retry.
+			break
 		}
 
-		// Either no successes, or no errors.
-		// No need to retry.
-		break
-	}
+		if !errors.As(err, &ie) {
+			// It's some other kind of error and we should return it.
+			return err
+		}
 
-	var skippedS []string
-	for resource := range skipped {
-		skippedS = append(skippedS, resource)
+		// It's an importError.
+		failuresTotal = ie.errMsgs
+
+		// If no successes, break out of the loop.
+		if len(successes) <= 0 {
+			break
+		}
+
+		// Some resources imported successfully, but others didn't.
+		// Time to retry.
+		log.Println("Some imports succeeded but others did not. Retrying the import, in case dependent values have now been populated.")
 	}
 
 	buf, err := template.WriteBuffer(summaryTmpl, map[string]interface{}{
-		"total":     len(successesTotal) + len(skippedS) + len(failuresTotal),
+		"total":     len(successesTotal) + len(skipped) + len(failuresTotal),
 		"successes": successesTotal,
-		"skipped":   skippedS,
+		"skipped":   skipped,
 		"failures":  failuresTotal,
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


xingao267:
What's a sample output from this?

internal/tfimport/tfimport.go
```diff
@@ -560,7 +581,7 @@ func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
 // If it imported some resources but failed to import others, it will return true for retry. This is a simple way to solve dependencies without having to figure out the graph.
 // A specific case: GKE node pool name depends on random_id; import the random_id first, then do the cycle again and import the node pool.
 // skipped is a map to be filled with skipped resources so they are skipped on subsequent runs too.
-func planAndImport(rn, importRn runner.Runner, runArgs *RunArgs, skipped map[string]bool) (retry bool, err error) {
+func planAndImport(rn, importRn runner.Runner, runArgs *RunArgs, skipped map[string]bool) (successes, errs []string, err error) {
```
umairidris, line 584: why do we need both errs and err?

umairidris:
If this is too much work feel free to ignore, but I think it would be really clean to hide all logs etc and just show a summary at the end:

Sample:

Found 10 unimported resources.

Imported 6 resources:
- google_project
- google_project_service["foo"]
...

Skipped 2 resources:
- google_project_lien
- ...

Failed to import 2 resources:
- google_service_account: reason for error
- ...

Then in -verbose mode it will show detailed logs etc. Or maybe we can use https://github.com/google/glog and just log at different verbosities.

internal/tfimport/tfimport.go
MartinPetkov: I think of it as:
* errs is errors from terraform import
* err is an error in the operation of the importer itself that prevents it from continuing

umairidris: In that case consider creating a custom error type and return that.

You can use errors.As and fmt.Errorf with %w to check and wrap the error as needed

https://blog.golang.org/go1.13-errors

MartinPetkov: I don't really want to treat errs as an error though. It's information for printing at the end, and may not get used at all if several retries ends up successfully importing all resources.

umairidris: How about returning a single error struct to differentiate between the two?

e.g.
type ImportError struct {
  Fatal  error
  Warnings []error
}

MartinPetkov: It still doesn't seem right to me. It may be more accurate to call it failedImports rather than errs.

umairidris: In general:

- when you call a func and it returns nil error it means that everything was ok. Thus, I don't know if failedImports conveys that message
- if the caller wants to differentiate between different failures and handle some different than others, then idiomatic way to do is to return specific error types which can be handled by the caller independently.
- in cases where the caller does not care about differentiating then they can simply log/return the error

It's one of the reasons why go has errors as regular types instead of exceptions in languages like java etc, the caller is supposed to handle the error if there is something non trivial about it.

At the end of the day this is an internal func so consequences are not too high. If you want to improve it with different var naming + more documentation that's fine, but it would likely get flagged in a readability review, especially if a public func.

MartinPetkov: Thank you, I've thought about this more and decide to do that, so I'm now using a custom error in the main retry loop.

MartinPetkov:
> If this is too much work feel free to ignore, but I think it would be really clean to hide all logs etc and just show a summary at the end:
> 
> Sample:
> 
> ```
> Found 10 unimported resources.
> 
> Imported 6 resources:
> - google_project
> - google_project_service["foo"]
> ...
> 
> Skipped 2 resources:
> - google_project_lien
> - ...
> 
> Failed to import 2 resources:
> - google_service_account: reason for error
> - ...
> ```
> 
> Then in -verbose mode it will show detailed logs etc. Or maybe we can use https://github.com/google/glog and just log at different verbosities.

Done. Used a template for this.

internal/tfimport/tfimport.go
```diff
@@ -517,6 +518,40 @@ func DoesNotExist(output string) bool {
...
+func (e *importErr) Error() string {
+	return fmt.Sprintf("failed to find or import %v resources. Use --verbose flag to see detailed error messages.", len(e.errMsgs))
+}
+
+var summaryTmpl string = `
+Summary:
+
+Found {{.total}} unimported resources.
+
+Imported {{len .successes}}{{if gt (len .successes) 0}}:{{else}}.{{end}}
```
xingao267, line 539: Successfully imported ...

internal/tfimport/tfimport.go
```diff
@@ -517,6 +518,40 @@ func DoesNotExist(output string) bool {
...
+}
+
+func (e *importErr) Error() string {
+	return fmt.Sprintf("failed to find or import %v resources. Use --verbose flag to see detailed error messages.", len(e.errMsgs))
+}
+
+var summaryTmpl string = `
+Summary:
+
+Found {{.total}} unimported resources.
```
xingao267, line 537: Found {{.total}} resources or importable resources i guess?

internal/tfimport/tfimport.go
```diff
@@ -517,6 +518,40 @@ func DoesNotExist(output string) bool {
 	return reDoesNotExist.FindStringIndex(output) != nil
 }
 
+// importErr is an error indicating a resource failed to be imported.
+// This doesn't indicate an error in running the steps around trying an import,
+// it's specifically for terraform failing to import a resource for a reason
+// that can't be handled (i.e. does not exist or not importable at all).
+type importErr struct {
```
umairidris, line 525: nit: don't trim struct names. 

importError

internal/tfimport/tfimport.go
```diff
@@ -538,19 +573,52 @@ func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
...
-			return err
+			if errors.As(err, &ie) {
+				failuresTotal = ie.errMsgs
+				if len(successes) > 0 {
+					// Some resources imported successfully, but others didn't.
+					// Time to retry.
+					log.Println("Some imports succeeded but others did not. Retrying the import, in case dependent values have now been populated.")
+					continue
+				}
+			} else {
```
umairidris, line 593: reduce indentation by putting simpler case first e.g.

if !errors.As(err, &ie) {
  return err
}

if len(successes) > 0 {
  ...
}

internal/tfimport/tfimport.go
```diff
@@ -538,19 +573,52 @@ func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
 		return fmt.Errorf("expand path %q: %v", inputDir, err)
 	}
 
+	var successesTotal, failuresTotal []string
 	skipped := make(map[string]bool)
+	var ie *importErr
```
umairidris, line 578: Can you just use ie.ErrMsgs instead of failuresTotal?

internal/tfimport/tfimport.go
```diff
@@ -538,19 +573,52 @@ func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
...
+			}
 		}
 
-		// If all imports succeeded, or none did, stop trying to import things, since we aren't making progress anymore.
-		if !retry {
-			break
-		}
+		// Either no successes, or no errors.
+		// No need to retry.
+		break
```
umairidris, line 601: ditto, you might be able to do

if err == nil {
  break
}

.. handle more complicated error case here ...

internal/tfimport/tfimport.go
```diff
@@ -538,19 +573,52 @@ func Run(rn runner.Runner, importRn runner.Runner, runArgs *RunArgs) error {
...
+
+	var skippedS []string
+	for resource := range skipped {
+		skippedS = append(skippedS, resource)
+	}
+
+	buf, err := template.WriteBuffer(summaryTmpl, map[string]interface{}{
+		"total":     len(successesTotal) + len(skippedS) + len(failuresTotal),
+		"successes": successesTotal,
+		"skipped":   skippedS,
```
umairidris, line 612: Why not pass skipped as a map to the template?

internal/tfimport/tfimport.go
```diff
@@ -656,28 +723,29 @@ func planAndImport(rn, importRn runner.Runner, runArgs *RunArgs, skipped map[str
...
 			skipped[cc.Address] = true
 
 		// Check if the error indicates insufficient information.
-		case errors.As(err, &ie):
-			log.Printf("Insufficient information to import %q\n", cc.Address)
+		case errors.As(err, &iie):
+			log.Printf("Insufficient information to import %q, missing fields %s\n", cc.Address, strings.Join(iie.MissingFields, ","))
 
-			errMsg := fmt.Sprintf("%q (insufficient information)", cc.Address)
+			failedMsg := fmt.Sprintf("%v (insufficient information)", cc.Address)
```
umairidris, line 744: i would revert these var renames. failedMsg just seems like a longer way of saying errMsg

umairidris (approved):

internal/tfimport/tfimport.go
MartinPetkov: No, since ie may be nil.

internal/tfimport/tfimport.go
MartinPetkov: I thought it wouldn't work, but I tried it and it does, thanks!

PR #798, opened 2021-02-12
minor change in usage log of import.sh

xingao267:
from feedback in tg/985938

### DIFF
--- a/examples/tfengine/generated/team/cicd/configs/import.sh
+++ b/examples/tfengine/generated/team/cicd/configs/import.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -ex
+set -e
 
 DIRS=()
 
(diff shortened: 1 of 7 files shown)


examples/tfengine/generated/team/cicd/configs/import.sh
```diff
@@ -23,7 +23,7 @@ do
   case $c in
     d) DIRS="${OPTARG}" ;;
     *)
-      echo "Invalid flag ${OPTARG}"
```
MartinPetkov, line 26: Why not print both?

MartinPetkov (approved):

examples/tfengine/generated/team/cicd/configs/import.sh
xingao267: It's already the option is illegal like below

```
$ ./import.sh -b
./import.sh: illegal option -- b
Usage: import.sh -d="dir1 dir2 ..."
``

umairidris (approved):

Issue #462, opened 2020-07-23
Detect unmanaged files in output dir


umairidris:
The output dir could be non-empty, or have old generated files that were not cleaned up. Make that visible to the user.

MartinPetkov:
This should be controllable by flag.

umairidris:
I was thinking this would just be output (not an exit status non zero). Did you have something else in mind?

MartinPetkov:
That would be fine. I would like this: by default it just warns that you generated files in a non-empty directory, and if you set a flag, it would delete files in output directories before writing out the new ones.

xingao267:
Wanted by solutions team to have an optional flag to clean up the output_dir. Default to false to not break current behavior.

xingao267:
@MartinPetkov is it something you have bandwidth to take on? An optional flag to clean up the output dir is sufficient for now. Default to false. A special case is devops, which we don't want to cleanup as it might contain local state during initial deployment when state has not been backed up yet.

umairidris:
Deleting files can lead to some unexpected behaviour, I would suggest just outputting a list of extra files/dirs. The user should be able to take the list and just pass it to rm if they want I imagine...

MartinPetkov:
Yes, I think this is something I can take on. I'll assign myself and start looking.

xingao267:
or it could be a enum flag for different mode

xingao267:
cc @jordan83

xingao267:
> Deleting files can lead to some unexpected behaviour, I would suggest just outputting a list of extra files/dirs. The user should be able to take the list and just pass it to rm if they want I imagine...

what about always listing out existing files/dirs, but have a mode to force delete?

umairidris:
Sure, I just want to avoid the case where user is used to setting a delete flag and then there is a typo/missed dir and they nuke their home dir or something :)

xingao267:
Yea, good point. Probably just outputting a list of extra files/dirs is enough to serve the purpose and avoid potential disastrous behavior.

PR #823, opened 2021-03-01
Add tfengine flag to show unmanaged output files

MartinPetkov:
Does not halt execution, simply shows those files.

Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/462


cmd/tfengine/main.go
```diff
@@ -31,12 +31,13 @@ import (
...
-	format      = flag.Bool("format", true, "Whether to format generated files.")
-	templates   = flag.String("templates", "", "Comma-separated list of templates to generate. Leave empty for all.")
-	addLicenses = flag.Bool("add_licenses", true, "Whether to add license headers to generated Terraform files.")
-	showVersion = flag.Bool("version", false, "show version and exit")
+	configPath    = flag.String("config_path", "", "Path to config file")
+	outputPath    = flag.String("output_path", "", "Path to directory dump output")
+	format        = flag.Bool("format", true, "Whether to format generated files.")
+	templates     = flag.String("templates", "", "Comma-separated list of templates to generate. Leave empty for all.")
+	addLicenses   = flag.Bool("add_licenses", true, "Whether to add license headers to generated Terraform files.")
+	showUnmanaged = flag.Bool("show_unmanaged_files", false, "Whether to show unmanaged files in the output directory that were not generated by tfengine. Will not delete them.")
```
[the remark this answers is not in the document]
MartinPetkov: Done.

internal/tfengine/tfengine.go
```diff
@@ -272,3 +279,55 @@ func ConvertToLocalBackend(path string) error {
...
+			if !deleteFiles {
+				log.Printf("WARNING: Unmanaged file: %v\n", path)
+				return nil
+			}
+
+			fullPath, err := fileutil.Expand(path)
+			if err != nil {
+				return err
+			}
+			log.Printf("WARNING: Deleting unmanaged file: %v\n", path)
```
xingao267, line 315: In a situation where the files are generated successfully once or were present in the output directory before, and then another tfengine run failed, with the current logic, it would list all the previously generated files from the output directory as unmanaged files (although they were actually managed files) iiuc. Is there a way to prevent that or avoid that confusion?

MartinPetkov: From what I can tell from the code, it won't get to this step if it failed to generate the files, it would return an error early. The only time it would get errors and still look for unmanaged files is if adding licenses or formatting fails, but none of those will affect the file names, so it shouldn't print all files as if they were unmanaged.

Are you able to produce a case where it lists all previously-generated files?

internal/tfengine/tfengine.go
```diff
@@ -272,3 +279,55 @@ func ConvertToLocalBackend(path string) error {
...
+			os.RemoveAll(fullPath)
+
+			return nil
```
xingao267, line 318: return os.RemoveAll(fullPath)

internal/tfengine/tfengine.go
xingao267: You are right. Shouldn't get to this step if failed to generated files.

internal/tfengine/tfengine.go
```diff
@@ -272,3 +279,55 @@ func ConvertToLocalBackend(path string) error {
...
+				managed[trimmedPath] = true
+				return nil
+			}
+
+			if _, ok := managed[trimmedPath]; ok {
+				return nil
+			}
+
+			if !deleteFiles {
+				log.Printf("WARNING: Unmanaged file: %v\n", path)
```
xingao267, line 307: Just INFO should be enough, instead of WARNING

xingao267 (approved):

PR #866, opened 2021-04-01
Importer: Treat unimportable and not exists as skipped

MartinPetkov:


internal/tfimport/tfimport.go
```diff
@@ -761,8 +761,10 @@ func planAndImport(rn, importRn runner.Runner, runArgs *RunArgs, skipped map[str
 		// err will be `exit code 1` even when it failed because the resource is not importable or already exists.
 		case NotImportable(output):
 			log.Printf("Import not supported by provider for resource %q\n", cc.Address)
+			skipped[cc.Address] = true
```
xingao267, line 764: With this change, these resources will be listed under `Skipped` instead of `Failed to import`?

MartinPetkov: Currently they're not listed at all.

xingao267 (approved):

xingao267:
Do you want to cut a release and use it in the solution templates?

MartinPetkov:
> Do you want to cut a release and use it in the solution templates?

No, not needed right now, it's just something I observed and wanted to update.

PR #868, opened 2021-04-07
Add Terraform Engine integration tests

### LINKED ISSUE #852: Add a basic integration test to deployment team.hcl at presubmit

xingao267:
Fixes #852

The `integration-tests` build is configured to be only triggered when a repository collaborator or owner comments `/gcbrun`.

### DIFF
--- a/build/presubmit_int.yaml
+++ b/build/presubmit_int.yaml
@@ -20,7 +20,7 @@ timeout: 3600s
 
 options:
   env: ["GOPATH=/go"]
-  machineType: "N1_HIGHCPU_32"
+  machineType: "N1_HIGHCPU_8"
 
 steps:
 - name: "${_CFT_TOOLS_CONTAINER}"
@@ -36,9 +36,10 @@ steps:
   id: Terraform Engine integration tests
   env:
   - "RUN_INTEGRATION_TEST=true"
-  - "BILLING_ACCOUNT=$_BILLING_ACCOUNT"
-  - "FOLDER_ID=$_FOLDER_ID"
-  - "DOMAIN=$_DOMAIN"
+  # These values are configured in the Cloud Build Trigger as substitutions.
+  - "BILLING_ACCOUNT=${_BILLING_ACCOUNT}"
+  - "FOLDER_ID=${_FOLDER_ID}"
+  - "DOMAIN=${_DOMAIN}"
 
 - name: "${_CFT_TOOLS_CONTAINER}"
   entrypoint: "bash"
--- a/build/presubmit_int.yaml
+++ b/build/presubmit_int.yaml
@@ -1,0 +1,47 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Terraform version used is 0.14.8.
+substitutions:
+  _CFT_TOOLS_CONTAINER: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:8b0d85766044b0e4701729f9e70ee78dcecdb8688bd86915eb2a1cb00580ba1e"
+
+timeout: 3600s
+
+options:
+  env: ["GOPATH=/go"]
+  machineType: "N1_HIGHCPU_32"
+
+steps:
+- name: "${_CFT_TOOLS_CONTAINER}"
+  entrypoint: "terraform"
+  args: ["version"]
+  waitFor: ["-"]
+  id: Terraform version
+
+- name: "${_CFT_TOOLS_CONTAINER}"
+  entrypoint: "go"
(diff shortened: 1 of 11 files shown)


examples/tfengine/modules/team.hcl
```diff
@@ -103,13 +103,18 @@ template "project_networks" {
...
+        display_name = "Bastion Accessor Service Account"
+      }]
       bastion_hosts = [{
         name           = "bastion-vm"
         network        = "$${module.network.network.network.self_link}"
         subnet         = "$${module.network.subnets[\"{{.default_location}}/bastion-subnet\"].self_link}"
         image_family   = "ubuntu-2004-lts"
         image_project  = "ubuntu-os-cloud"
-        members        = ["group:{{.prefix}}-bastion-accessors@{{.domain}}"]
+        members        = ["serviceAccount:$${google_service_account.bastion_accessor.email}"] # Placeholder for testing.
```
MartinPetkov, line 117: How come this needs to be a serviceAccount and not a group?

examples/tfengine/modules/team.hcl
```diff
@@ -382,13 +391,10 @@ EOF
...
           }
         }]
-        retention_policy = {
-          retention_period = 86400 # 1 day.
-        }
-        iam_members = [{
-          role   = "roles/storage.objectViewer"
-          member = "group:{{.prefix}}-data-viewers@{{.domain}}"
-        }]
+        # Skip for integration tests.
```
MartinPetkov, line 394: Nit: I would rephrase slightly: "This is skipped for integration tests for <reason> [I assume so resources can be destroyed after], uncomment when writing a production template."

tests/tfengine/integration_test.go
```diff
@@ -17,8 +17,11 @@
 //  - The environment variable RUN_INTEGRATION_TEST must be set to "true".
 //  - The environment variables BILLING_ACCOUNT and FOLDER_ID must be set.
```
MartinPetkov, line 18: And DOMAIN must also be set

MartinPetkov (approved):
Overall LGTM.

Just one suggestion, if the destroy fails I think we'd need to destroy the resources manually/debug. It would be good to also add either a README for how to do this manually or a manually-triggered GCB job to do cleanup (or both), which I can imagine would be "1) Generate configs the same way as the integration tests do 2) Run importer 3) Run destroy again".

This is unlikely to happen though so I'm fine just leaving it for the future. I'll open an issue.

MartinPetkov:
Oh, also, I'm wondering, will this integration test clash if multiple PRs are opened at the same time?

MartinPetkov:
> Oh, also, I'm wondering, will this integration test clash if multiple PRs are opened at the same time?

I looked again and I think this line prevents this from happening:
```
prefix := fmt.Sprintf("dpt%v", time.Now().Unix())
```

build/presubmit_int.yaml
```diff
@@ -0,0 +1,47 @@
...
+  id: Terraform version
+
+- name: "${_CFT_TOOLS_CONTAINER}"
+  entrypoint: "go"
+  args: ["test", "./tests/tfengine/...", "-v", "-timeout", "60m"]
+  waitFor: ["-"]
+  id: Terraform Engine integration tests
+  env:
+  - "RUN_INTEGRATION_TEST=true"
+  - "BILLING_ACCOUNT=$_BILLING_ACCOUNT"
```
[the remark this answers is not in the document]
xingao267: Done

build/presubmit_int.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+# Terraform version used is 0.14.8.
+substitutions:
+  _CFT_TOOLS_CONTAINER: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:8b0d85766044b0e4701729f9e70ee78dcecdb8688bd86915eb2a1cb00580ba1e"
+
+timeout: 3600s
+
+options:
+  env: ["GOPATH=/go"]
+  machineType: "N1_HIGHCPU_32"
```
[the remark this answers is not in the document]
xingao267: Let me try with `N1_HIGHCPU_8` and report back

examples/tfengine/modules/team.hcl
xingao267: because otherwise we need to supply an existing group in the test. Using a service account created on the fly to avoid that. It's also how CFT's tests do it for IAM members.

examples/tfengine/modules/team.hcl
xingao267: Done

tests/tfengine/integration_test.go
xingao267: Done

xingao267:
> Overall LGTM.
> 
> Just one suggestion, if the destroy fails I think we'd need to destroy the resources manually/debug. It would be good to also add either a README for how to do this manually or a manually-triggered GCB job to do cleanup (or both), which I can imagine would be "1) Generate configs the same way as the integration tests do 2) Run importer 3) Run destroy again".
> 
> This is unlikely to happen though so I'm fine just leaving it for the future. I'll open an issue.

Good point. That can happen. We can simply delete the projects in the testing folder manually for now.

build/presubmit_int.yaml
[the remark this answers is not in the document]
xingao267: NI_HIGHCPU_8 is enough. Almost no difference in time.

MartinPetkov:
Just FYI the integration tests take 36 min 47 sec with just 8 CPUs. Seems like it's better to keep it with the higher resources, which seemed to take much less time even if it may take longer for the build to start.

xingao267:
with 32, it takes about the same time.

PR #876, opened 2021-04-14
Update go.sum

MartinPetkov:
I just ran `go mod tidy`


MartinPetkov:
I have some local pre-commit hooks that wouldn't run unless I ran `go mod tidy`, but I didn't want to do both in one PR.

xingao267 (approved):

PR #879, opened 2021-04-19
do not init backend during validate and lock state in plan

### LINKED ISSUE #878: Validate build should not access state
If both validate and plan are accessing the state file at the same time, one could fail to get the lock and error out. The `terraform init` run in validate should not access the state.

xingao267:
Fixes #878


xingao267:
Tested CICD using our testing projects.

MartinPetkov (approved):
To confirm:
* Validate won't access the state file at all.
* Plan will access it but won't lock it, since it's not modifying it.
* Apply will access it and lock it.

xingao267:
> To confirm:
> 
> * Validate won't access the state file at all.
> * Plan will access it but won't lock it, since it's not modifying it.
> * Apply will access it and lock it.

That's right. I added this summary to the README.

PR #916, opened 2021-06-24
feat: Update schema docs structure

ernestognw:
## Changes

- Change schema docs to table form
- Added fields to schema
- Edit `generate_schema_docs` script to use the new schema fields
- Added `default` and `Required` to docs


xingao267:
You can send a PR against master branch for this.

scripts/generate_schema_docs/template.go
```diff
@@ -24,23 +24,9 @@ var tmpl = template.Must(template.New("").Parse(`# {{.Title}}
...
-{{- if .Type}}
-
-Type: {{.Type}}
-{{- end}}
-
-{{- if .Pattern}}
-
-Pattern: {{.Pattern}}
-{{- end}}
+| {{$name}} | {{or .Description "-"}} | {{or .Type "-"}} | {{.RequiredByParent}} | {{or .Default "-"}} | {{or .Pattern "-"}} |
```
MartinPetkov, line 30: Should the fourth column be `{{or .RequiredByParent .Required}}`?

MartinPetkov:
The test is failing due to  markdownlint failing (see `./build/md_check.sh`. I can suggest the following:

* Disable the line check by adding a file `markdownlint_config.json` with the following, and updating the md_check.sh script to point to the file with `-c markdownlint_config.json`:

```
{
    "line-length": false
}
```
   (see https://github.com/DavidAnson/markdownlint#tags)

* For `MD010/no-hard-tabs`, don't use tabs in `scripts/generate_schema_docs/template.go`. The lines don't actually have to be aligned in the .md file.
* For `MD022/blanks-around-headings/blanks-around-headers`, add a line after `## Properties`.
* For `MD033/no-inline-html`, maybe use two spaces or `\\`? If only `<br>` works, then let's just ignore that rule too.
* For `MD034/no-bare-urls`, investigate why the final `>` of the URL in templates/tfengine/recipes/resources.hcl isn't getting added to the .md file.

docs/tfengine/schemas/audit.md
```diff
@@ -3,102 +3,19 @@
...
-Pattern: ^[a-z][a-z0-9\-]{4,28}[a-z0-9]$
-
-### storage_location
-
-Location of logs storage bucket.
-
-Type: string
+| Property 	| Description 						| Type 	   			   | Required			   		   | Default             | Pattern 			 			 |
+| --------- | ----------------------- | ---------------- | --------------------- | ------------------- | ------------------- |
+| additional_filters | Additional filters for log collection and export. List entries will be        concatenated by "OR" operator. Refer to        <https://cloud.google.com/logging/docs/view/query-library> for query syntax.        Need to escape \ and " to preserve them in the final filter strings.        See example usages under "examples/tfengine/".        Logs with filter "logName:\"logs/cloudaudit.googleapis.com\"" is always enabled. | array(string) | false | - | - |
```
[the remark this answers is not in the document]
ernestognw: Done

scripts/generate_schema_docs/template.go
ernestognw: I don't think so. The `Required` field is at parent level and it's an array of keys. Is converted to boolean flags for their properties while executing `generate_schema_docs`. If it's in `Required, it should be `RequiredByParent` at the next level

scripts/generate_schema_docs/main.go
```diff
@@ -118,13 +118,38 @@ func schemaFromHCL(b []byte) (*schema, error) {
...
+		switch prop.Type {
+		case "object":
+			addRequiredByParent(prop.Properties, requiredToMap(prop.Required))
+		case "array":
+			addRequiredByParent(prop.Items.Properties, requiredToMap(prop.Items.Required))
+		}
+	}
+}
+
+func requiredToMap(required []string) map[string]bool {
```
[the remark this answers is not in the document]
ernestognw: Done. It doesn't appear because comment was added to a previous commit. It's on the `Files Changed` tab

docs/tfengine/schemas/project.md
```diff
@@ -3,128 +3,24 @@
...
+| parent_id | ID of parent GCP resource to apply the policy        Can be one of the organization ID or folder ID according to parent_type. | string | false | - | - |
+| parent_type | Type of parent GCP resource to apply the policy        Can be one of 'organization' or 'folder'. | string | false | - | ^organization|folder$ |
+| project | Config for the project. | object | true | - | - |
+| project.api_identities | The list of service identities (Google Managed service account for the API) to            force-create for the project (e.g. in order to grant additional roles).            APIs in this list will automatically be appended to `apis`.            Not including the API in this list will follow the default behaviour for identity            creation (which is usually when the first resource using the API is created).            Any roles (e.g. service agent role) must be explicitly listed.            See <https://cloud.google.com/iam/docs/understanding-roles#service-agent-roles-roles>            for a list of related roles. | array(object) | false | - | - |
+| project.api_identities.api | - | string | false | - | - |
```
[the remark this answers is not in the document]
ernestognw: It doesn't seems to have one. [Look](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/project.hcl#L85)

Do you suggest any?

docs/tfengine/schemas/audit.md
```diff
@@ -3,102 +3,19 @@
...
+| bigquery_location | Location of logs bigquery dataset. | string | false | - | - |
+| logs_bigquery_dataset | Bigquery Dataset to host audit logs for 1 year. Useful for querying recent activity. | object | false | - | - |
+| logs_bigquery_dataset.dataset_id | ID of Bigquery Dataset. | string | false | - | - |
+| logs_bigquery_dataset.sink_name | Name of the logs sink, default to "bigquery-audit-logs-sink". | string | false | - | - |
+| logs_storage_bucket | GCS bucket to host audit logs for 7 years. Useful for HIPAA audit log retention requirements. | object | false | - | - |
+| logs_storage_bucket.name | Name of GCS bucket. | string | false | - | - |
+| logs_storage_bucket.sink_name | Name of the logs sink, default to "storage-audit-logs-sink". | string | false | - | - |
+| parent_id | ID of parent GCP resource to apply the policy.        Can be one of the organization ID or folder ID according to parent_type. | string | false | - | ^[0-9]{8,25}$ |
+| parent_type | Type of parent GCP resource to apply the policy.        Must be one of 'organization' or 'folder'." | string | false | - | ^organization|folder$ |
```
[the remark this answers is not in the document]
ernestognw: Nice catch. For now I'm just escaping it in `Pattern` field, since it is not anywhere else. But in the case that is needed for description or other field too, it's pretty straightforward to fix

docs/tfengine/schemas/project.md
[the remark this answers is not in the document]
MartinPetkov: I suggest adding a description to the field in project.hcl. Something like:

```
description = "The API to apply the roles to. Example: healthcare.googleapis.com"
```

ernestognw:
> The test is failing due to markdownlint failing (see `./build/md_check.sh`. I can suggest the following:
> 
> * Disable the line check by adding a file `markdownlint_config.json` with the following, and updating the md_check.sh script to point to the file with `-c markdownlint_config.json`:
> 
> ```
> {
>     "line-length": false
> }
> ```
> 
> (see https://github.com/DavidAnson/markdownlint#tags)
> 
> * For `MD010/no-hard-tabs`, don't use tabs in `scripts/generate_schema_docs/template.go`. The lines don't actually have to be aligned in the .md file.
> * For `MD022/blanks-around-headings/blanks-around-headers`, add a line after `## Properties`.
> * For `MD033/no-inline-html`, maybe use two spaces or `\\`? If only `<br>` works, then let's just ignore that rule too.
> * For `MD034/no-bare-urls`, investigate why the final `>` of the URL in templates/tfengine/recipes/resources.hcl isn't getting added to the .md file.

Fix, this is what happened:

* I created a `.markdownlint.json` as markdownlint suggested, so it's automatically readed
* I removed the hard tabs from the template
* Added space after `## Propertires`
* For `MD034/no-bare-urls`, it was happening because URLs should end with a point like this:

```
"See <https://foo.bar/anywhere>."
```

Previously it was:
```
"See <https://foo.bar/anywhere>"
```

Somehow the `>"` is ignore, but I don't even know how to Google about this.

In general, whenever this happens, the fix is just to add the point at the very end.

MartinPetkov:
> > The test is failing due to markdownlint failing (see `./build/md_check.sh`. I can suggest the following:
> > 
> > * Disable the line check by adding a file `markdownlint_config.json` with the following, and updating the md_check.sh script to point to the file with `-c markdownlint_config.json`:
> > 
> > ```
> > {
> >     "line-length": false
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (see https://github.com/DavidAnson/markdownlint#tags)
> > 
> > * For `MD010/no-hard-tabs`, don't use tabs in `scripts/generate_schema_docs/template.go`. The lines don't actually have to be aligned in the .md file.
> > * For `MD022/blanks-around-headings/blanks-around-headers`, add a line after `## Properties`.
> > * For `MD033/no-inline-html`, maybe use two spaces or `\\`? If only `<br>` works, then let's just ignore that rule too.
> > * For `MD034/no-bare-urls`, investigate why the final `>` of the URL in templates/tfengine/recipes/resources.hcl isn't getting added to the .md file.
> 
> Fix, this is what happened:
> 
> * I created a `.markdownlint.json` as markdownlint suggested, so it's automatically readed
> * Line length check was disabled _only_ for tables
> * I removed the hard tabs from the template
> * Added space after `## Propertires`
> * For `MD034/no-bare-urls`, it was happening because URLs should end with a point like this:
> 
> ```
> "See <https://foo.bar/anywhere>."
> ```
> 
> Previously it was:
> 
> ```
> "See <https://foo.bar/anywhere>"
> ```
> 
> Somehow the `>"` is ignored, but I don't even know how to Google about this.
> 
> In general, whenever this happens, the fix is just to add the point at the very end.

Awesome, thanks for the extra context!

MartinPetkov (approved):

templates/tfengine/recipes/project.hcl
```diff
@@ -84,6 +84,7 @@ schema = {
             properties = {
               api = {
                 type = "string"
+                description = "The API to apply the roles to. Example: healthcare.googleapis.com"
```
xingao267, line 87: This should be "The API whose default Service Agent will be force-created and granted the role. Example: healthcare.googleapis.com."

You can update it in a separate PR.

templates/tfengine/recipes/audit.hcl
```diff
@@ -101,7 +101,7 @@ schema = {
         <https://cloud.google.com/logging/docs/view/query-library> for query syntax.
         Need to escape \ and " to preserve them in the final filter strings.
         See example usages under "examples/tfengine/".
-        Logs with filter "logName:\"logs/cloudaudit.googleapis.com\"" is always enabled.
+        Logs with filter `"logName:\"logs/cloudaudit.googleapis.com\""` is always enabled.
```
xingao267, line 104: why do we need to add ` here? If ` is added here, then the backspash might need to be removed. If you didn't test it with Cloud Logging, maybe keep it as before.

ernestognw: The complete paragraph explains that it is an example of how to escape quotes `"`. Afaik it's intended to be escaped since that's the example.

Am I wrong?

ernestognw: Update: You're right! I misunderstood. I'll update it with the description suggested in the other comment

xingao267: After looking at https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/examples/tfengine/multi_envs.hcl#L194, i think both are wrong. We should follow the pattern there, so that the generated configs will be like https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/examples/tfengine/generated/multi_envs/audit/main.tf#L77

ernestognw: This particular line is the final result at `.tf` file. I think it's good as it is right now, since it's just specifying the output of the config. 

Agree that we should follow https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/examples/tfengine/multi_envs.hcl#L194 while writing configs, but that's already specified at the field description.

I'm a little confused haha

xingao267: Resolved offline. The misunderstanding is from we are looking at different things. I was looking at the raw md file, and Ernesto was looking at the rich text. All good.

PR #919, opened 2021-06-24
fix: Change project recipe project.api_identities.api description

ernestognw:
## Changes

- Update description to field added at #916

### DIFF
--- a/templates/tfengine/recipes/project.hcl
+++ b/templates/tfengine/recipes/project.hcl
@@ -84,7 +84,7 @@ schema = {
             properties = {
               api = {
                 type = "string"
-                description = "The API to apply the roles to. Example: healthcare.googleapis.com"
+                description = "The API whose default Service Agent will be force-created and granted the role. Example: healthcare.googleapis.com."
               }
               roles = {
                 description = "Roles to granted to the API Service Agent."
--- a/templates/tfengine/recipes/project.hcl
+++ b/templates/tfengine/recipes/project.hcl
@@ -84,7 +84,7 @@ schema = {
             properties = {
               api = {
                 type = "string"
-                description = "The API whose default Service Agent will be force-created and granted the role. Example: healthcare.googleapis.com."
+                description = "The API whose default Service Agent will be force-created and granted the roles. Example: healthcare.googleapis.com."
               }
               roles = {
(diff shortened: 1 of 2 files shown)


templates/tfengine/recipes/project.hcl
```diff
@@ -87,7 +87,7 @@ schema = {
                 description = "The API to apply the roles to. Example: healthcare.googleapis.com"
               }
               roles = {
-                description = "Roles to granted to the API Service Agent."
+                description = "The API whose default Service Agent will be force-created and granted the role. Example: healthcare.googleapis.com."
```
xingao267, line 90: Should be updated on the `api` field.

MartinPetkov:
Thanks, it seems I misunderstood what "roles" actually did. Another reason why it was worth adding a description to this field.

templates/tfengine/recipes/project.hcl
ernestognw: You're right. I'm so sorry, didn't catch it D:

templates/tfengine/recipes/project.hcl
```diff
@@ -84,7 +84,7 @@ schema = {
             properties = {
               api = {
                 type = "string"
-                description = "The API to apply the roles to. Example: healthcare.googleapis.com"
+                description = "The API whose default Service Agent will be force-created and granted the role. Example: healthcare.googleapis.com."
```
xingao267, line 87: nit: s/role/roles

MartinPetkov (commented):
LGTM but I'll let Xin give the final Approval

templates/tfengine/recipes/project.hcl
ernestognw: DId you mean `... granted the role/roles`?

xingao267: `granted the roles.`

with the additional s because I noticed the roles field is a list.

We usually use `s/before/after` as a shorthand for replacement. It's from sed command syntax. Sorry for the confusion.

ernestognw: Ohh got it. Sure, np! Done

xingao267 (approved):

xingao267:
Thanks @ernestognw !

PR #928, opened 2021-07-02
Update terraform (major)

xingao267:


xingao267:
After looking more closely, it has several breaking changes including storage bucket https://github.com/terraform-google-modules/terraform-google-cloud-storage/blob/master/docs/upgrading_to_v2.0.md. Will close this PR and only upgrade necessary modules to avoid a major breaking update like this.

PR #947, opened 2021-07-09
Improve docs around data map overriding precedence.

xingao267:


docs/tfengine/template_dev.md
```diff
@@ -78,17 +78,78 @@ developers.
...
-1. Data blocks in templates follow a Tree structure. Values in the `data` block
-    from an upper level template are passed down to its child templates and made
-    available. There is no need to repeat those values in the `data` block in
-    child templates unless you would like to override them. For example, all
-    values specified in the top level template
-    [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/dd2464e1a293b0bc549ef34d0c73787e798e88ba/examples/tfengine/team.hcl#L17)
-    are passed down and made available to child templates
-    [root.hcl](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/dd2464e1a293b0bc549ef34d0c73787e798e88ba/examples/tfengine/modules/root.hcl),
-    [foundation.hcl](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/dd2464e1a293b0bc549ef34d0c73787e798e88ba/examples/tfengine/modules/foundation.hcl)
+1. `data` maps can be specified either
```
MartinPetkov, line 81: There's a typo above in "uniqie"

docs/tfengine/template_dev.md
```diff
@@ -78,17 +78,78 @@ developers.
...
+    or
+    [outside](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/c27ddc3f1b629f77a70e7b53908882a7e4c94f42/examples/tfengine/modules/team.hcl#L15)
+    a `template` block.
+
+    In both cases, values in the `data` maps from an upper level template are
+    passed down to its child template and made available. There is no need to
+    repeat data values in child templates unless you would like to override
+    them. For example, all data values specified in the top level template
+    [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/c27ddc3f1b629f77a70e7b53908882a7e4c94f42/examples/tfengine/team.hcl#L17)
+    are passed down and made available to its child templates
```
MartinPetkov, line 92: Nit: Mention "its transitive child templates" since team.hcl doesn't directly call foundation.hcl and team.hcl.

docs/tfengine/template_dev.md
```diff
@@ -78,17 +78,78 @@ developers.
...
+    [root.hcl](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/c27ddc3f1b629f77a70e7b53908882a7e4c94f42/examples/tfengine/modules/root.hcl),
+    [foundation.hcl](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/c27ddc3f1b629f77a70e7b53908882a7e4c94f42/examples/tfengine/modules/foundation.hcl)
     and
-    [team.hcl](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/dd2464e1a293b0bc549ef34d0c73787e798e88ba/examples/tfengine/modules/team.hcl).
+    [team.hcl](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/c27ddc3f1b629f77a70e7b53908882a7e4c94f42/examples/tfengine/modules/team.hcl).
+
+    However, in the two cases, the `data` maps' value overriding precedence are
```
MartinPetkov, line 98: Nit: "precedence is"

docs/tfengine/template_dev.md
```diff
@@ -78,17 +78,78 @@ developers.
...
+    different, which follows the 3 rules below:
+
+    1. Values spcified in the `data` maps **inside** the `template` block take
```
MartinPetkov, line 101: Typo: "spcified"

MartinPetkov: Should mention that this sub-rule is just within the same file, since the next two sub-rules define how data maps in child and parent files interact.

docs/tfengine/template_dev.md
```diff
@@ -78,17 +78,78 @@ developers.
...
+
+    ```hcl
+    # modules/bigquery.hcl
+    resource "google_bigquery_dataset" "dataset" {
+      dataset_id                  = "example_dataset"
+      location                    = "{{.bigquery_location}}"
+    }
+    ```
+
+    The 4 locations specified will have the following precedence, from high to
```
MartinPetkov, line 150: IIUC, this precedence is because:
* Values defined in the local template block (in root.hcl) take highest precedence ("D").
* Values provided by "parent" files (i.e. main.hcl) take precedence over ones defined in local global data blocks (i.e. "C")
* In the parent file, values defined in the template block ("B") take precedence over ones defined in the parent's global data block ("A").
* So the root.hcl file finds the value by the following: "use D since it's directly specified; but if it wasn't there, check if the parent has provided a value; if it hasn't, or there's no parent, use C". And in the parent, it does: "use B since it's directly specified; but if it wasn't, check the parent; if it hasn't, or there's no parent, use A" and passes that down to root.hcl

I'm sure my explanation above can be simplified, but I feel like there should be at least another sentence explaining why this particular order in this example.

xingao267: Added some explanation against the 3 rules. PTAL.

docs/tfengine/template_dev.md
MartinPetkov: Ah I see.

docs/tfengine/template_dev.md
MartinPetkov: Awesome, I like the way you explained it. Thank you!

MartinPetkov (approved):

PR #948, opened 2021-07-12
feat: Apply BP3 spec to audit component recipe

ernestognw:
## Changes

- Improve `terraform.vars` and `variables.tf` to audit recipe
- Updated default and required variables at config level 

Solves #913


docs/tfengine/schemas/audit.md
```diff
@@ -7,14 +7,14 @@
...
+| logs_bigquery_dataset | Bigquery Dataset to host audit logs for 1 year. Useful for querying recent activity. | object | true | - | - |
+| logs_bigquery_dataset.dataset_id | ID of Bigquery Dataset. | string | true | - | - |
+| logs_bigquery_dataset.sink_name | Name of the logs sink. | string | false | bigquery-audit-logs-sink | - |
+| logs_storage_bucket | GCS bucket to host audit logs for 7 years. Useful for HIPAA audit log retention requirements. | object | true | - | - |
+| logs_storage_bucket.name | Name of GCS bucket. | string | true | - | - |
+| logs_storage_bucket.sink_name | Name of the logs sink. | string | false | storage-audit-logs-sink | - |
 | parent_id | ID of parent GCP resource to apply the policy.        Can be one of the organization ID or folder ID according to parent_type. | string | false | - | ^[0-9]{8,25}$ |
 | parent_type | Type of parent GCP resource to apply the policy.        Must be one of 'organization' or 'folder'." | string | false | - | ^organization\|folder$ |
 | project | Config of project to host auditing resources | object | false | - | - |
```
ernestognw, line 20: I didn't see these `project` variables used in the audit recipe. I suppose they're inherited from `project` recipe somehow, since there is a `module.project.project_id` at the `main.tf`

Two questions:

1. If it's always inherited, I suppose it's here to override the default since it's not required, isn't?
2. I saw some other inherited variables that are not pretty clear when they're required or not. Most of them seem required since it's templating syntax is like they were, but I'm not sure if I have to add them as required.

examples/tfengine/generated/org_foundation/audit/variables.tf
```diff
@@ -13,9 +13,46 @@
...
 
 variable "auditors_group" {
-  type = string
+  type        = string
+  description = "This group will be granted viewer access to the audit log dataset and bucket as well as security reviewer permission on the root resource specified."
+}
+
+variable "filter" {
+  type        = list(string)
+  description = "Filters for log collection and export concatenated by \"OR\" operator. Refer to https://cloud.google.com/logging/docs/view/query-library for query syntax."
```
ernestognw, line 31: Last PR i was talking with @MartinPetkov that there are cases when the description doesn't completely match with the variable. This is another example.

The original variable name at config level is `additional_filters`, and this is is `filter`, which is just a combination of defaults and the `additional`.

Nothing to change, just wanted to point it out.

examples/tfengine/generated/folder_foundation/audit/terraform.tfvars
```diff
@@ -13,5 +13,16 @@
 # limitations under the License.
 
 
-folder         = "folders/12345678"
-auditors_group = "example-auditors@example.com"
+folder            = "folders/12345678"
+auditors_group    = "example-auditors@example.com"
+filter            = "logName:\"logs/cloudaudit.googleapis.com\" OR logName=\"logs/forseti\" OR logName=\"logs/application\""
```
MartinPetkov, line 18: Minor nit, I find it hard to parse with all these backslash escapes, would it be possible to replace it with a heredoc string? As in
```
filter = <<-EOF
logName:"logs/cloudaudit.googleapis.com" OR logName="logs/forseti" OR logName="logs/application"
EOF
```

templates/tfengine/components/audit/variables.tf
```diff
@@ -15,13 +15,55 @@
...
 }
 {{- end}}
 
 variable "auditors_group" {
   type = string
+  description = "This group will be granted viewer access to the audit log dataset and bucket as well as security reviewer permission on the root resource specified."
+}
+
+variable "filter" {
+  type = list(string)
```
MartinPetkov, line 41: It seems like it should just be `string`, not `list(string)`.

MartinPetkov (approved):

examples/tfengine/generated/org_foundation/audit/variables.tf
ernestognw: Might be solved now that #953 is done

templates/tfengine/components/audit/variables.tf
ernestognw: That's correct. Fixed. Thanks

examples/tfengine/generated/folder_foundation/audit/terraform.tfvars
ernestognw: I don't think it's possible because they have to land the config that way 'cause it is passed to be passed with the `"` escaped, but I think I maybe lack of context.

Wait to see what @xingao267 says

MartinPetkov: If I'm understanding correctly the escapes are just because of HCL syntax using double quotes for strings. In the API itself I don't think it needs escapes, it just takes the string with the quotes. So I think using EOF heredoc is equivalent to using quotes with escapes.

examples/tfengine/generated/org_foundation/audit/main.tf
```diff
@@ -71,9 +71,9 @@ module "bigquery_export" {
   source  = "terraform-google-modules/log-export/google"
   version = "~> 6.0.0"
 
-  log_sink_name          = "bigquery-audit-logs-sink"
+  log_sink_name          = var.logs_bigquery_dataset.sink_name
   destination_uri        = module.bigquery_destination.destination_uri
-  filter                 = "logName:\"logs/cloudaudit.googleapis.com\""
+  filter                 = "logName:\"logs/cloudaudit.googleapis.com\"${var.additional_filters}"
```
MartinPetkov, line 76: I would suggest making `additional_filters` a `list(string)` and using the [join](https://www.terraform.io/docs/language/functions/join.html) function to concatenate them with the correct separator. Otherwise someone could add the filter incorrectly or without a space in the beginning for example.

examples/tfengine/generated/folder_foundation/audit/terraform.tfvars
ernestognw: Tbh I'm not sure. The schema says that it should be kept, [look](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/docs/tfengine/schemas/audit.md).

Is this somehow tested by the unit/integration tests? I can save some time if that's true, otherwise I think I'll have to test it manually.

examples/tfengine/generated/org_foundation/audit/main.tf
ernestognw: Wonderful. Thanks!

examples/tfengine/generated/folder_foundation/audit/terraform.tfvars
MartinPetkov: I think our docs are subtly wrong here. The escapes are just for Terraform, not for the API. Looking at the sample queries in https://cloud.google.com/logging/docs/view/query-library, it looks like the API expects normal quotes, not escaped quotes, and we're only escaping them since we're wrapping them in `""`. Using EOF/heredocs would make it more obvious what the final filter string is.

MartinPetkov: From what I can tell it's not tested by the integration tests. They don't set up any modules which use additional_filters.

ernestognw: I applied the changes, but I'm not sure if we're in the same situation as when this discussion appeared, 'cause now the `additional_filters` is a `list(string)` instead of `string`. Therefor, the `EOF` thing doesn't apply.

Anyway, I did [this commit](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/948/commits/84097ff3272c67f1ca4f079bb1a20a794191bf0a) with what I think is correct according to what we talked

Let me know if I'm misunderstanding something, so I can revert the commit and change it

ernestognw: Nevermind. Test are not passing because of it.

Actually, now that it is a `list(string)`, I think we're good to go as it was previously.

Just confirm and I'll remove that commit. We can discuss it in our meeting

MartinPetkov: Yes, I think with `list(string)` it should be left as it was. In the recipes we need to use double escapes because during the conversion to .tf files strings get escaped, so in .hcl we write `\\\"`so that it becomes `\"` in .tf.

ernestognw: Got it.
I'm resolving and reverting the commit

templates/tfengine/components/audit/variables.tf
```diff
@@ -15,13 +15,62 @@
...
+  type = object({
+    name = string
+    sink_name = string
+  })
+  description = "GCS bucket to host audit logs for 7 years. Useful for HIPAA audit log retention requirements."
+}
+
+variable "storage_location" {
+  type = string
+  description = "	Location of logs storage bucket."
```
xingao267, line 75: extra space in the description

examples/tfengine/generated/org_foundation/audit/main.tf
```diff
@@ -95,9 +95,9 @@ module "storage_export" {
   source  = "terraform-google-modules/log-export/google"
   version = "~> 6.0.0"
 
-  log_sink_name          = "storage-audit-logs-sink"
+  log_sink_name          = var.logs_storage_bucket.sink_name
   destination_uri        = module.storage_destination.destination_uri
-  filter                 = "logName:\"logs/cloudaudit.googleapis.com\""
+  filter                 = join(" OR ", concat(["logName:\"logs/cloudaudit.googleapis.com\""], var.additional_filters))
```
xingao267, line 100: have you tried deployed and tested it?

xingao267 (approved):

templates/tfengine/components/audit/variables.tf
ernestognw: Fixed. Thanks!

examples/tfengine/generated/org_foundation/audit/main.tf
ernestognw: Not yet. I assumed tests cared about that. Let me try it.

Is there any particular recommendation on what I'm searching for? I'm not sure if everything can be deployed correctly with a malformed filter string without logged errors.

xingao267: The integration tests don't deploy the audit recipe i believe. The unit tests might run a local terraform plan on the audit recipe. Can you try deploy the audit recipe in our testing org and pass in some additional filter strings, then verify if the filter is set properly by referring to log export doc here https://cloud.google.com/logging/docs/export/configure_export_v2

ernestognw: I deployed an audit example and created these two sinks:

Screenshot 2021-07-15 1 05 08 PM

filters: `logName:"logs/cloudaudit.googleapis.com" OR logName="logs/forseti" OR logName="logs/application"`

Screenshot 2021-07-15 1 04 44 PM

filters: `logName:"logs/cloudaudit.googleapis.com" OR logName="logs/forseti" OR logName="logs/application"`

ernestognw: So far, everything seems to be ok. I'll leave the config for a few hours so you can check it out before destroying it.

examples/tfengine/generated/multi_envs/audit/terraform.tfvars
```diff
@@ -15,3 +15,17 @@
 
 folder         = "folders/12345678"
 auditors_group = "example-auditors@example.com"
+additional_filters = [
+  "logName=\"logs/forseti\"",
```
xingao267, line 19: let's change the = signs to : to be consistent with the default one, i.e. logName:\"logs/cloudaudit.googleapis.com\"

ernestognw: Done. I also did `terraform apply` with the new change and everything seems to be fine

PR #961, opened 2021-07-16
Simplify CICD triggers

### LINKED ISSUE #956: Avoid repeated resources on cicd recipe
## Description

Due to #949, `triggers.tf` was refactored to be submodules. It turns out that these modules use a lot of repeated logic and can be simplified into a single trigger that uses a different configuration `filename`

Started by [this discussion](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/949#issuecomment-880128723)

ernestognw:
## Changes

As per https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/949#pullrequestreview-706577998, triggers were simplified to avoid repetition

- Created a generic `trigger` module
- Changed `validate` and `plan` triggers to use this generic trigger

Closes #956


ernestognw (commented):
I was able to simplify `validate` and `plan` triggers, but still have small problems with the `apply` trigger. So for now, it was left as it is, but I'd like to discuss two blockers to simplify this one:

1. [This line](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/components/cicd/templates/tfengine/components/cicd/triggers/apply/main.tf#L31) at the `apply` trigger is not the same as [this line](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/961/files#diff-f2ff83b1896da49607a20f551fcfeff5f7573efd73ce3963fa7f330002f14ad7R30) in this PR, and is just that line the one that makes it uncompatible with the generic `trigger`. 

Any idea? I don't find a way to conditionally specify it from the module declaration. And [the docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudbuild_trigger) specify that `push` and `pull_request` are mutually exclusive so there is no way to put both and disable one. 

Also, templating here is not possible because the generic `trigger` is only one file. We need a variation of that file depending on the parent `triggers` module. (I'm doing my best to explain this, we can have a GVC if needed)

2. [This description](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/components/cicd/templates/tfengine/components/cicd/triggers/apply/main.tf#L21) is incompatible with a generalization of the `trigger` module. This is fixable if we agree to provide a description for all three of them. The problem is that the generic `trigger` has an `on_push` trigger and a `scheduled` trigger, but the `apply` trigger seems to use the same one for both cases. I know this depend on the config `.yaml` file used, but there are several ways to fix this. I'll try to update if I find any proposal.

ernestognw:
> 1. [This line](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/components/cicd/templates/tfengine/components/cicd/triggers/apply/main.tf#L31) at the `apply` trigger is not the same as [this line](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/961/files#diff-f2ff83b1896da49607a20f551fcfeff5f7573efd73ce3963fa7f330002f14ad7R30) in this PR, and is just that line the one that makes it uncompatible with the generic `trigger`.
> 
> Any idea? I don't find a way to conditionally specify it from the module declaration. And [the docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudbuild_trigger) specify that `push` and `pull_request` are mutually exclusive so there is no way to put both and disable one.
> 
> Also, templating here is not possible because the generic `trigger` is only one file. We need a variation of that file depending on the parent `triggers` module. (I'm doing my best to explain this, we can have a GVC if needed)

Solved by creating a single trigger, and passing a dynamic block.
 
> 2. [This description](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/components/cicd/templates/tfengine/components/cicd/triggers/apply/main.tf#L21) is incompatible with a generalization of the `trigger` module. This is fixable if we agree to provide a description for all three of them. The problem is that the generic `trigger` has an `on_push` trigger and a `scheduled` trigger, but the `apply` trigger seems to use the same one for both cases. I know this depend on the config `.yaml` file used, but there are several ways to fix this. I'll try to update if I find any proposal.

Solved by creating a single trigger and passing the description

xingao267:
Feel free to close this one if no longer needed.

PR #963, opened 2021-07-19
Add automatic multiline detection and trimmed descriptions

### LINKED ISSUE #962: Trim schema descriptions for tfengine rendering
## Description

Now that #953 has been added, schema info is available for templating, but in order to render them, there are two following concerns regarding `description` fields:

1. Many of the descriptions are multi-line descriptions rendered with heredoc syntax, and they inherit some extra spaces from the original schema, so, when using in templates, there are some unusual spaces and linebreaks on the final result.

For example, from a schema:

```
...
  auditors_group = {
      description = <<EOF
        This group will be granted viewer access to the audit log dataset and
        bucket as well as security reviewer permission on the root resource
        specified.
      EOF
      type        = "string"
    }
...
```

When rendering, it's set like this:

```
variable "auditors_group" {
  description = <<EOF

           This group will be granted viewer access to the audit log dataset and
           bucket as well as security reviewer permission on the root resource
           specified.
  EOF
}
```

Note how extra tabs/line breaks might be there
 
2. Since descriptions could be heredoc or single line, it is needed a way to render according to what's needed depending on the string. This avoids to double-check how description is written in schema each time.

For example, a description like this:

```
description = "A description".
```

Should be automatically rendered between quotes since it's single line. On the other hand:

```
description = <<EOF
  This is
  a multiline
  description.
EOF
```

Should be rendered with heredoc in `variables.tf`

ernestognw:
## Changes

- Trim schema descriptions for `tfengine` usage
- Add `schemaDescription` function to use at `variables.tf` to automatic render singleline or heredoc

Fixes #962


internal/template/funcmap.go
```diff
@@ -145,3 +146,13 @@ func regexReplaceAll(regex string, s string, repl string) (string, error) {
 func makeSlice(args ...interface{}) []interface{} {
 	return args
 }
+
+// schemaDescription returns a heredoc or single line string
+// according to the description provided format
+func schemaDescription(s string) string {
+	if strings.Contains(s, "\n") {
+		return fmt.Sprintf("<<EOF\n%s\nEOF", s)
```
MartinPetkov, line 154: Maybe add one more \n at the end since the final EOF needs to be alone on its own line.

internal/tfengine/tfengine.go
```diff
@@ -336,3 +338,34 @@ func findUnmanaged(generatedDir, outputDir string, deleteFiles bool) error {
...
+				description := v.Interface().(string)
+				trimmedDescription := strings.TrimSpace(leftTrim(description))
+				schema["description"] = trimmedDescription
+			}
+		}
+	}
+}
+
+// leftTrim trims leading space from every line.
+func leftTrim(s string) string {
```
MartinPetkov, line 364: Function looks fine but what is the purpose of removing leading spaces from every line?

MartinPetkov: Oh I realized there's an issue linked. I'll resolve this while I read that.

internal/tfengine/tfengine.go
```diff
@@ -336,3 +338,34 @@ func findUnmanaged(generatedDir, outputDir string, deleteFiles bool) error {
...
+		v := val.MapIndex(key)
+
+		switch t := v.Interface().(type) {
+		case map[string]interface{}:
+			trimDescriptions(t)
+		case string:
+			if key.String() == "description" {
+				description := v.Interface().(string)
+				trimmedDescription := strings.TrimSpace(leftTrim(description))
+				schema["description"] = trimmedDescription
```
MartinPetkov, line 357: Is this the correct map access? Shouldn't it be something like `v[key] = trimmedDescription`?

internal/template/funcmap.go
ernestognw: Makes sense but that final `\n` is added on templating. I'll do my best to explain:

This is an example of how it's used:

```
1- ...
2- description = {{schemaDescription __schema__.properties.someProperty.description}}
3- ...
```

What you're suggesting is the natural linebreak that we intentionally put between line 2 and 3 in the template.

Here's an example of the output so you can validate everything is alright:

https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/964/files#diff-1d81e567868c331d72f2628f1287d393e95b80ccba397396aed80c84eb1292e2R19-R22

MartinPetkov: Oh right. Templating whitespace is a bit weird since you can do `{{-` and `-}}` to eat whitespace, but by default it adds a newline because (as you said) we naturally add one as part of the template. Thanks for the details, resolving.

internal/tfengine/tfengine.go
ernestognw: The problem is that v is a reflection, which is not the actual schema map. I'm not too familiar with the concept of reflections and I had trouble writing on it. The alternative would be to add the `struct types` for `schema` and `properties` so we can parse it and traverse more accurately, but I considered an overengineered solution.

Let me know what you think

MartinPetkov: I was more asking if this is correct, I don't think you need to switch to struct types. It just doesn't seem obvious to me, but does it work?

ernestognw: Sure, it works.

I decided to send #965 and #964 to show examples of how it shows up.

My only concern would be for deep schema descriptions. The only nested property I had used is [this one](**url**)

Not sure if I can wait to use it to test on deep descriptions such as in `cicd` recipe. But I guess it's fine.

MartinPetkov: I think the "this one" link is broken but I see. LGTM then

MartinPetkov (approved):

internal/tfengine/tfengine.go
ernestognw: You're right. My bad. This is the link if you want to take a look:

https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/964/files#diff-e73c9c3e385a38ef46a5b55bef489f61a47403acd9fcfd49728325209d48c2deR80

I'm resolving

internal/template/funcmap.go
```diff
@@ -145,3 +146,13 @@ func regexReplaceAll(regex string, s string, repl string) (string, error) {
 func makeSlice(args ...interface{}) []interface{} {
 	return args
 }
+
+// schemaDescription returns a heredoc or single line string
+// according to the description provided format
```
xingao267, line 151: period(.) at the end

internal/tfengine/tfengine.go
```diff
@@ -336,3 +338,34 @@ func findUnmanaged(generatedDir, outputDir string, deleteFiles bool) error {
 
 	return nil
 }
+
+// trimDescriptions remove tabs and extra space from schema descriptions
+// so they can be used during rendering without unexpected gaps between lines
```
xingao267, line 343: period at the end

internal/template/funcmap.go
```diff
@@ -24,17 +24,18 @@ import (
...
+	"hcl":               hcl,
+	"hclField":          hclField,
+	"merge":             merge,
+	"replace":           replace,
+	"resourceName":      resourceName,
+	"now":               time.Now,
+	"trimSpace":         strings.TrimSpace,
+	"regexReplaceAll":   regexReplaceAll,
+	"makeSlice":         makeSlice,
+	"schemaDescription": schemaDescription,
```
xingao267, line 38: where is this function used?

xingao267: Are you planning to use it in following PRs in recipe/component templates?

xingao267:
LGTM. Thanks! Few minor nits. And also should this change also be merged to the main branch?

xingao267 (approved):

internal/template/funcmap.go
ernestognw: I just decided to separate this minor update from the actual implementations. There are two PRs using this.

Look at [this example ](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/965/files#diff-302c27f4dbc9649902ac53af0fb0be28dc03d8bb5883d76a98a27743c8d71fddR19)

xingao267: Got it. Thanks.

ernestognw:
> LGTM. Thanks! Few minor nits. And also should this change also be merged to the main branch?

Done, pointed to main branch. Thanks!

xingao267:
seems it's mixed with other commits at the main branch...

ernestognw:
> seems it's mixed with other commits at the main branch...

Noted right after pointing to master. FIxed. Thanks!

PR #964, opened 2021-07-19
Components/devops use schema

### LINKED ISSUE #957: Apply schema variables into devops recipe template
## Description

As a result from #953, the schema variables are now available at templating, which enables the project to maintain a single source of information.

The data that can be inserted into the templates is:
- Default values
- Descriptions
- Regex validations

This issue aims to implement `__schema__` variable to devops recipe

ernestognw:
## Changes

- Implemented `__schema__` usage on devops recipe template

Note: Review depends on #963 

Closes #957


docs/tfengine/schemas/devops.md
```diff
@@ -15,7 +15,7 @@
...
 | billing_account | ID of billing account to attach to this project. | string | false | - | - |
 | enable_gcs_backend | Whether to enable GCS backend for the devops module.        Defaults to false.<br><br>Since the devops module creates the state bucket, it cannot back up        the state to the GCS bucket on the first module. Thus, this field        should be set to false initially.<br><br>After the devops module has been applied once and the state bucket        exists, the user should set this to true and regenerate the configs.<br><br>To migrate the state from local to GCS, run `terraform init` on the        module. | boolean | false | false | - |
-| parent_id | ID of parent GCP resource to apply the policy.        Can be one of the organization ID or folder ID according to parent_type. | string | false | - | ^[0-9]{8,25}$ |
+| parent_id | ID of parent GCP resource to apply the policy.        Can be one of the organization ID or folder ID according to parent_type.        See <https://cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy> to learn more about resource hierarchy. | string | false | - | ^[0-9]{8,25}$ |
```
xingao267, line 18: just wondering is it possible to remove the extra spaces between the sentences? not a big deal as it doesn't show the extra spaces when rendering in markdown

examples/tfengine/generated/devops/devops/variables.tf
```diff
@@ -16,7 +16,10 @@ variable "admins_group" {
   type = object({
     id = string
   })
-  description = "Group which will be given admin access to the folder or organization."
+  description = <<EOF
+Group which will be given admin access to the folder or organization.
+It will be created if 'exists' is false.
```
xingao267, line 21: This sentence is confusing as the `exists` is not a var here. Maybe remove it from the schema description as well.

templates/tfengine/recipes/devops.hcl
```diff
@@ -32,6 +32,7 @@ schema = {
       description = <<EOF
         ID of parent GCP resource to apply the policy.
         Can be one of the organization ID or folder ID according to parent_type.
+        See <https://cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy> to learn more about resource hierarchy.
```
xingao267, line 35: At line 162 and 77 of this file (the descriptions for exists for two groups), you can say

"Whether or not the group exists already. It will be created it not."

docs/tfengine/schemas/devops.md
ernestognw: Sure, let me see if I can do it easily and I'll open another PR to master (doing it here would be a lot of overhead for all the files). I agree it's not a big deal since it doesn't really create a problem.

examples/tfengine/generated/devops/devops/variables.tf
ernestognw: This is something I mentioned with @MartinPetkov on my very first Pull Request. There are descriptions that should be slightly different from the rendered version, but we agreed to keep it like that.

We can go deep on that discussion here. Personally, I think it can be removed for this case since the `exists` schema property is already documented, but there might be other cases where we shouldn't remove something from the schema.

xingao267: Yea, I think removing in this case is probably fine. In other cases, we might need to develop some other mechanism to include only part of the description from the schema.

ernestognw: Great, removed it and updated.

I'm resolving

xingao267 (approved):

ernestognw:
Forgot to regen. Sorry

/gcbrun

docs/tfengine/schemas/devops.md
ernestognw: Solved by in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/969

ernestognw:
DNS zone conflicted due to parallel integrations test run.

/gcbrun

MartinPetkov (approved):

PR #965, opened 2021-07-19
Components/audit use schema

### LINKED ISSUE #959: Apply schema variables into audit recipe template
## Description

As a result from #953, the schema variables are now available at templating, which enables the project to maintain a single source of information.

The data that can be inserted into the templates is:
- Default values
- Descriptions
- Regex validations

This issue aims to implement `__schema__` variable to audit recipe

ernestognw:
## Changes

- Implemented `__schema__` usage on audit recipe template

Note: Review depends on #963 

Closes #959

### DIFF
--- a/examples/tfengine/generated/folder_foundation/audit/variables.tf
+++ b/examples/tfengine/generated/folder_foundation/audit/variables.tf
@@ -14,10 +14,7 @@
 
 variable "folder" {
   type        = string
-  description = <<EOF
-ID of parent GCP resource to apply the policy.
-Can be one of the organization ID or folder ID according to parent_type.
-EOF
+  description = "ID of the parent GCP resource to apply the configuration."
   validation {
     condition     = can(regex("^folders/[0-9]{8,25}$", var.folder))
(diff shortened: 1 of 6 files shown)


examples/tfengine/generated/folder_foundation/audit/variables.tf
```diff
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 variable "folder" {
   type        = string
-  description = "ID of the folder to apply the configuration."
+  description = <<EOF
+ID of parent GCP resource to apply the policy.
+Can be one of the organization ID or folder ID according to parent_type.
```
xingao267, line 19: It will be confusing as the variable is named `folder`.

Maybe change the schema description to something more generic like

"ID of the parent GCP resource to apply the configuration."

(Note should be configuration instead of policy at the end)

ernestognw: Maybe later we can explore some sort of interpolation or something like that to replace `parent` with `folder` or `organization`. I'm just throwing a moonshot.

For now, I left it as you requested. Resolving.

xingao267 (approved):

PR #972, opened 2021-07-22
fix: components/cicd issues

ernestognw:
## Description

Tested CICD deployment and found some typos and issues with variables. Those are fixed in this PR

### DIFF
--- a/examples/tfengine/generated/multi_envs/cicd/terraform.tfvars
+++ b/examples/tfengine/generated/multi_envs/cicd/terraform.tfvars
@@ -66,7 +66,7 @@ envs = [
       }
       plan = {
         skip            = true
-        run_on_push     = true
+        run_on_push     = false
         run_on_schedule = ""
       }
       apply = {
--- a/examples/tfengine/generated/multi_envs/cicd/terraform.tfvars
+++ b/examples/tfengine/generated/multi_envs/cicd/terraform.tfvars
@@ -22,7 +22,7 @@ build_editors         = ["group:example-cicd-editors@example.com"]
 
 build_viewers = ["group:example-cicd-viewers@example.com"]
 
-cloud_source_repostory = {
+cloud_source_repository = {
   name = "example"
   readers = [
     "group:example-source-readers@example.com",
@@ -65,7 +65,9 @@ envs = [
         run_on_schedule = ""
       }
       plan = {
-        skip = true
+        skip            = true
+        run_on_push     = true
+        run_on_schedule = ""
       }
       apply = {
         skip            = false
--- a/examples/tfengine/generated/org_foundation/cicd/main.tf
+++ b/examples/tfengine/generated/org_foundation/cicd/main.tf
@@ -176,7 +176,7 @@ resource "google_organization_iam_member" "cloudbuild_sa_organization_iam" {
 
 # Create Google Cloud Build triggers for specified environments
 module "triggers" {
-  for_each = var.envs
+  for_each = { for env in var.envs : env.name => env }
   // TODO(ernestognw): Merge triggers to simplify resources #956
   source = "./triggers"
 
@@ -190,7 +190,6 @@ module "triggers" {
   // TODO(ernestognw): Look how to calculate terraform_root_prefix from terraform_root
   terraform_root        = var.terraform_root
   terraform_root_prefix = var.terraform_root_prefix
-  service_account_email = google_service_account.cloudbuild_scheduler_sa.email
 
   depends_on = [
     google_project_service.services,
--- a/examples/tfengine/generated/team/cicd/main.tf
+++ b/examples/tfengine/generated/team/cicd/main.tf
@@ -174,7 +174,7 @@ resource "google_folder_iam_member" "cloudbuild_sa_folder_iam" {
 
 # Create Google Cloud Build triggers for specified environments
 module "triggers" {
-  for_each = var.envs
+  for_each = { for env in var.envs : env.name => env }
   // TODO(ernestognw): Merge triggers to simplify resources #956
   source = "./triggers"
 
@@ -188,7 +188,6 @@ module "triggers" {
   // TODO(ernestognw): Look how to calculate terraform_root_prefix from terraform_root
   terraform_root        = var.terraform_root
   terraform_root_prefix = var.terraform_root_prefix
-  service_account_email = google_service_account.cloudbuild_scheduler_sa.email
 
   depends_on = [
     google_project_service.services,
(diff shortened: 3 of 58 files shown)


templates/tfengine/components/cicd/triggers/variables.tf
```diff
@@ -112,4 +112,5 @@ variable "terraform_root_prefix" {
 variable "service_account_email" {
   type = string
   description = "Email of the Cloud Scheduler service account."
+  default = ""
```
MartinPetkov, line 115: Why is this optional?

templates/tfengine/recipes/cicd.hcl
```diff
@@ -168,11 +168,6 @@ schema = {
 
             NOTE: The modules will be deployed in the given order. If a module
             depends on another module, it should show up after it in this list.
-
-            NOTE: The CICD has permission to update APIs within its own project.
```
MartinPetkov, line 172: I missed the previous discussion, how come this paragraph is no longer needed?

xingao267: It's not a commonly used feature, and would actually cause more problems. I removed it from the example in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/819 but didn't remove it from the schema.

templates/tfengine/components/cicd/triggers/variables.tf
ernestognw: It's because the block that defines the resource that fills the variable is optional:

```
resource "google_service_account" "cloudbuild_scheduler_sa" {
  project      = var.project_id
  account_id   = "cloudbuild-scheduler-sa"
  display_name = "Cloud Build scheduler service account"
  depends_on = [
    google_project_service.services,
  ]
}
```

It only appears if there are scheduled jobs. Now it's needed to conditionally render this variable, son when it's not there, the default it's empty to avoid the error, but it's not really needed since it's not used.

MartinPetkov (approved):

examples/tfengine/generated/org_foundation/cicd/main.tf
```diff
@@ -176,7 +176,7 @@ resource "google_organization_iam_member" "cloudbuild_sa_organization_iam" {
 
 # Create Google Cloud Build triggers for specified environments
 module "triggers" {
-  for_each = var.envs
```
xingao267, line 179: It's a list of strings, you can use toset(var.envs) for this.

examples/tfengine/generated/team/cicd/main.tf
```diff
@@ -174,7 +174,7 @@ resource "google_folder_iam_member" "cloudbuild_sa_folder_iam" {
 
 # Create Google Cloud Build triggers for specified environments
 module "triggers" {
-  for_each = var.envs
```
xingao267, line 177: same here

templates/tfengine/components/cicd/terraform.tfvars
```diff
@@ -69,20 +69,29 @@ envs = [
...
           {{- end}}
         }
         plan = {
           skip = {{not (has .triggers "plan")}}
           {{- if has .triggers "plan"}}
           run_on_push = {{get .triggers.plan "run_on_push" true}}
           run_on_schedule = "{{get .triggers.plan "run_on_schedule" ""}}"
+          {{- else}}
+          run_on_push = true
+          run_on_schedule = ""
```
xingao267, line 84: why do we need these? Because those vars are mandatory? If so, the default for a bool variable should be set to `false`.

examples/tfengine/generated/org_foundation/cicd/main.tf
ernestognw: No, it's a list of objects. The error that gives me when used as `var.envs` is this:
Screenshot 2021-07-22 4 45 17 PM

And with `{for env in var.envs:  env.name => env}` is this:

Screenshot 2021-07-22 4 43 28 PM

templates/tfengine/components/cicd/terraform.tfvars
ernestognw: Yeah, the whole object is mandatory and avoiding subkeys result in an error. I decided to put `run_on_push` in true because that's the default specified on the schema.

Doesn't really matter since it is not used 'cause it's skipped but I'll update to false

examples/tfengine/generated/org_foundation/cicd/main.tf
xingao267: Got it. You're right. I thought it's a list of strings.

templates/tfengine/components/cicd/terraform.tfvars
ernestognw: Done

templates/tfengine/components/cicd/terraform.tfvars
```diff
@@ -69,20 +69,29 @@ envs = [
...
           {{- end}}
         }
         plan = {
           skip = {{not (has .triggers "plan")}}
           {{- if has .triggers "plan"}}
           run_on_push = {{get .triggers.plan "run_on_push" true}}
           run_on_schedule = "{{get .triggers.plan "run_on_schedule" ""}}"
+          {{- else}}
+          run_on_push = false
+          run_on_schedule = ""
```
xingao267, line 84: Actually i think you can probably do this so it's cleaner. could you try it?

```
          run_on_push = {{get .triggers "plan.run_on_push" true}}
          run_on_schedule = "{{get .triggers "plan.run_on_schedule" ""}}"
```

ernestognw: It worked. I ran `regen` and the output didn't changed.

Thanks

examples/tfengine/generated/multi_envs/cicd/terraform.tfvars
```diff
@@ -65,7 +65,9 @@ envs = [
         run_on_schedule = ""
       }
       plan = {
-        skip = true
+        skip            = true
+        run_on_push     = true
```
xingao267, line 69: Actually from looking at this generated config. I think it's less confusing if `run_on_push` is set to `false` when `skip` is set to true. Maybe reverting back to https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/972/commits/00a09cfbf08c09438ea1a8166b02ef38391363ce is a bit better. Sorry for the back and forth.

ernestognw: np, I agree.

I did a reset to 00a09cf

xingao267 (approved):

PR #978, opened 2021-07-29
Update releasing.md

pasha-gh:


RELEASING.md
```diff
@@ -46,7 +46,7 @@ latest versions of everything.
 
 ### Frequency
 
-We strive to make a release at least once a month, containing accumulated changes since the last release. We may make
+We strive to make a release as frequently as possible, containing accumulated changes since the last release. We may make
```
MartinPetkov, line 49: Maybe this whole sentence should be reworded to communicate that we release an unstable version as soon as possible after each change, and release major/minor/patch versions less frequently.

pasha-gh: Sure, is this the plan for the foreseeable future?

MartinPetkov: That's the plan right now. There's a GitHub Action that builds new binaries on commit but otherwise we release new template versions infrequently.

pasha-gh: How about the following:

We release an unstable version as soon as possible after each change. This release may include breaking changes. We may make versioned releases any time as needed. The MAJOR/MINOR/PATCH versions can be updated as needed.

MartinPetkov: LGTM just minor suggestions:
* Change "as soon as possible" to "after each commit via an automated release flow".
* Change "the major/minor/patch versions can be" to "are" updated as needed.

xingao267 (approved):

PR #984, opened 2021-08-02
feat: Reduce templating dependency on devops/main.tf

### LINKED ISSUE #981: Remove main.tf templating logic from Devops component
While generating `variables.tf` file transferred some of the major logic into the `terraform.tfvars` file, there were some /conditional resources and fields that are still dependant on templating.

As stated at the discussion of #957, there were 3 main problems with the complete reduction of templating logic:

1. **Conditional resources**: Some resources turned out to be dependant on the existence of some variables. This conditional can be reproduced with the [`count` meta argument](https://www.terraform.io/docs/language/meta-arguments/count.html)
2. **Conditional nested fields**: Conditional nested fields were kept with templated `if` statements due to the restriction that terraform has of omitting nested fields. This can be solved by providing default templated values on `.tfvars` that are the defaults on the underlying terraform modules (every module has a default.). Passing the underlying default is equivalent to not pass anything at all and it avoids the error caused by the nested variable omission.
3. **Descriptions for nested values**: Nested values were not with a description provided, but now, with #953, descriptions are now more maintainable since they're added as variables to the templating, and, now tested against internal resources, I'm proposing a format to show nested descriptions like the following:

```
variable "foo" {
  type = object({
    nestedFoo: object({
      nestedSubFoo: ...,
    }),
    nestedBar: ...
   })

  description = <<EOF
    {{props.foo.description}}

   * nestedFoo = {{...props.nestedFoo.description}}
   ** nestedSubFoo = {{...props.nestedSubFoo.description}}
   * nestedBar = {{...props.nestedFoo.description}}
  EOF
}
```

Applying these 3 principles to devops recipe, templating can be theoretically removed

ernestognw:
Closes #981


templates/tfengine/components/devops/terraform.tfvars
```diff
@@ -11,17 +11,16 @@
...
-  display_name = "{{get .admins_group "display_name" (regexReplaceAll "@.*" .admins_group.id "")}}"
-  customer_id  = "{{.admins_group.customer_id}}"
-  {{hclField .admins_group "description" -}}
-  {{hclField .admins_group "owners" -}}
-  {{hclField .admins_group "managers" -}}
-  {{hclField .admins_group "members" -}}
-  {{- end -}}
+  owners       = [{{range get . "admins_group.owners"}}"{{.}}",{{end}}]
+  managers       = [{{range get . "admins_group.managers"}}"{{.}}",{{end}}]
+  members       = [{{range get . "admins_group.members"}}"{{.}}",{{end}}]
```
MartinPetkov, line 23: For style, I think the one-liner version here is less readable than the expanded multi-line version, I would prefer you use the latter.

templates/tfengine/components/devops/variables.tf
```diff
@@ -12,28 +12,30 @@
...
-    {{- if has .admins_group "members"}}
     members      = list(string)
-    {{- end}}
-    {{- end}}
   })
-  description = {{schemaDescription $props.admins_group.description}}
+  description = <<EOF
+    {{$props.admins_group.description}}
+
+    * customer_id = {{$adminsGroupProps.customer_id.description}}
```
MartinPetkov, line 32: Same as the other PR, consider just looping through `$adminsGroupProps` here and adding a "Fields:" line right before to make it clearer what this section of the description is.

ernestognw: Same resolution as the other PR, only added Field header

templates/tfengine/components/devops/terraform.tfvars
ernestognw: You're right, I was kinda unsure wether to use this one or the multiline. For empty cases, the multiline ends up being like:

```
owners = [
]
```

I changed it to multiline

examples/tfengine/generated/org_foundation/devops/main.tf
```diff
@@ -56,16 +63,84 @@ module "state_bucket" {
   location   = var.storage_location
 }
 
+# Devops project owners group.
+module "owners_group" {
+  source  = "terraform-google-modules/group/google"
+  version = "~> 0.2"
+
+  count = var.project.owners_group.exists ? 1 : 0
```
xingao267, line 71: the logic is opposite.

count should 1 when the group does not exist

examples/tfengine/generated/org_foundation/devops/variables.tf
```diff
@@ -11,12 +11,31 @@
...
+    Fields:
+
+    * customer_id = Customer ID of the organization to create the group in.
+See <https://cloud.google.com/resource-manager/docs/organization-policy/restricting-domains#retrieving_customer_id>
+for how to obtain it.
+    * description = Description of the group.
+    * display_name = Display name of the group.
+    * exists = Whether or not the group exists already. It will be created if not.
+    * id = Email address of the group.
+    * owners = Owners of the group.
```
xingao267, line 37: what about managers and members?

templates/tfengine/components/devops/variables.tf
```diff
@@ -50,32 +54,45 @@ variable "parent_id" {
...
+
+    * apis = {{$projectProps.apis.description}}
+    * owners_group = {{$projectProps.owners_group.description}}
+    ** customer_id {{$projectOwnersGroupProps.customer_id.description}}
+    ** description {{$projectOwnersGroupProps.description.description}}
+    ** display_name {{$projectOwnersGroupProps.display_name.description}}
+    ** exists {{$projectOwnersGroupProps.exists.description}}
+    ** id {{$projectOwnersGroupProps.id.description}}
+    ** owners {{$projectOwnersGroupProps.owners.description}}s
+    * project_id {{$projectProps.project_id.description}}s
```
xingao267, line 94: extra s at the end

templates/tfengine/components/devops/variables.tf
```diff
@@ -50,32 +54,45 @@ variable "parent_id" {
...
+    Fields:
+
+    * apis = {{$projectProps.apis.description}}
+    * owners_group = {{$projectProps.owners_group.description}}
+    ** customer_id {{$projectOwnersGroupProps.customer_id.description}}
+    ** description {{$projectOwnersGroupProps.description.description}}
+    ** display_name {{$projectOwnersGroupProps.display_name.description}}
+    ** exists {{$projectOwnersGroupProps.exists.description}}
+    ** id {{$projectOwnersGroupProps.id.description}}
+    ** owners {{$projectOwnersGroupProps.owners.description}}s
```
xingao267, line 93: extra s at the end

templates/tfengine/components/devops/variables.tf
```diff
@@ -50,32 +54,45 @@ variable "parent_id" {
...
+    {{$props.project.description}}
+
+    Fields:
+
+    * apis = {{$projectProps.apis.description}}
+    * owners_group = {{$projectProps.owners_group.description}}
+    ** customer_id {{$projectOwnersGroupProps.customer_id.description}}
+    ** description {{$projectOwnersGroupProps.description.description}}
+    ** display_name {{$projectOwnersGroupProps.display_name.description}}
+    ** exists {{$projectOwnersGroupProps.exists.description}}
```
xingao267, line 91: Can we use extra indent instead of double *?

examples/tfengine/generated/team/devops/terraform.tfvars
```diff
@@ -29,7 +35,17 @@ project = {
     "secretmanager.googleapis.com",
   ]
   owners_group = {
-    id = "example-devops-owners@example.com"
+    customer_id  = ""
+    description  = ""
+    display_name = "example-devops-owners"
+    exists       = true
+    id           = "example-devops-owners@example.com"
+    owners = [
```
xingao267, line 43: Can this be formatted as `owners = []` ?

ernestognw: I mentioned here https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/984#discussion_r682769050

But I think that's preferrable. Other functions like hcl write the emails without the "

```
owners = [user@example.com]
```

instead of

```
owners = ["user@example.com"]
```

I didn't found other way around

templates/tfengine/components/devops/variables.tf
ernestognw: We can, but that might add extra complexity if we automate this by iterating the properties as Martin suggested in other PRs.

What do you think might be the best?

examples/tfengine/generated/org_foundation/devops/variables.tf
ernestognw: They're disabled at the schema, we agreed on the first PR to keep them in templates but leaving them commented on the schema.

I cannot access the description if it's disabled

examples/tfengine/generated/org_foundation/devops/main.tf
ernestognw: That's right. Thanks!

templates/tfengine/components/devops/variables.tf
MartinPetkov: Maybe this should be part of the schemaDescription function, to also output the internal field descriptions. Then you could iterate and use recursion (for nested objects) with normal code instead of templating code. Would that be complex to do?

ernestognw: Sounds doable and better, but the problems are almost the same:

1. Sometimes properties are inside properties key or items key (this is fixable if I understand better when it's one or the other)
2. Some schema variables may be renamed into the `variables.tf` or don't appear at all. How can we detect those?

MartinPetkov: Good point.

Then in the interest of not debating this forever, I'm okay with not iterating for now. Besides that, I agree that an extra indent would look better than 2 **s

examples/tfengine/generated/org_foundation/devops/variables.tf
xingao267: oh right. my bad.

examples/tfengine/generated/team/devops/terraform.tfvars
xingao267: I actually mean removing the line break if it's an empty list, 

like [], instead of [
]


but it's very minor.

templates/tfengine/components/devops/variables.tf
ernestognw: Got it. Looks better, thanks!

MartinPetkov (commented):
LGTM but I'll let Xin approve

examples/tfengine/generated/team/devops/terraform.tfvars
ernestognw: Resolved in favor of https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/984#:~:text=I%20mentioned%20here-,%23984%20(comment),-But%20I%20think

templates/tfengine/components/devops/main.tf
```diff
@@ -157,25 +128,25 @@ module "admins_group" {
 # The group is not ready for IAM bindings right after creation. Wait for
 # a while before it is used.
 resource "time_sleep" "admins_wait" {
+  count = var.admins_group.exists ? 1 : 0
```
xingao267, line 131: 0 : 1

templates/tfengine/components/devops/terraform.tfvars
```diff
@@ -11,17 +11,16 @@
...
-  {{if $missing_admins_group -}}
-  display_name = "{{get .admins_group "display_name" (regexReplaceAll "@.*" .admins_group.id "")}}"
-  customer_id  = "{{.admins_group.customer_id}}"
-  {{hclField .admins_group "description" -}}
-  {{hclField .admins_group "owners" -}}
-  {{hclField .admins_group "managers" -}}
-  {{hclField .admins_group "members" -}}
-  {{- end -}}
+  owners       = [{{range get . "admins_group.owners"}}"{{.}}",{{end}}]
+  managers       = [{{range get . "admins_group.managers"}}"{{.}}",{{end}}]
```
xingao267, line 22: one extra space before the `=`.

Also format this in the same way as the below owners_group. I think the owners_group one looks a bit better.

templates/tfengine/components/devops/main.tf
```diff
@@ -108,47 +91,35 @@ module "owners_group" {
...
 
-{{- if $missing_admins_group}}
-
-# Admins group for at {{.parent_type}} level.
+# Admins group at parent level.
 module "admins_group" {
   source  = "terraform-google-modules/group/google"
   version = "~> 0.2"
 
+  count = var.admins_group.exists ? 1 : 0
```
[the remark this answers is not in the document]
ernestognw: You're right. Thanks!

examples/tfengine/generated/multi_envs/devops/terraform.tfvars
```diff
@@ -26,10 +29,18 @@ project = {
...
-    id           = "example-devops-owners@example.com"
-    display_name = "example-devops-owners"
     customer_id  = "c12345678"
-    owners       = ["user1@example.com"]
+    description  = ""
+    display_name = "example-devops-owners"
+    exists       = false
+    id           = "example-devops-owners@example.com"
+    owners = [
+      "user1@example.com",
```
xingao267, line 38: This might not work with. The `,` at the end might cause issue

templates/tfengine/components/devops/terraform.tfvars
ernestognw: Done, I set it as the owners_group

examples/tfengine/generated/multi_envs/devops/terraform.tfvars
ernestognw: Tested with these 3 different combinations on a cicd recipe. They all work.

I'm resolving

Screenshot 2021-08-11 11 57 03 AM
Screenshot 2021-08-11 11 59 54 AM
Screenshot 2021-08-11 12 00 38 PM

xingao267 (approved):

xingao267:
Thanks. LGTM. There is some merging conflicts. Could you take a look and once they are resolved, i'll merge it.

MartinPetkov (approved):

xingao267 (approved):

Issue #951, opened 2021-07-12
deployment.hcl terraform_addons property validation not working


xingao267:
For example, at https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/f18accfb3cf392a877eabc8e4d7d29310467f146/examples/tfengine/modules/team.hcl#L51

If changing the `providers` to some other string, tfengine does not throw an error. But it should because in the deployment recipe where those fields are defined, [additionalProperties](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/f18accfb3cf392a877eabc8e4d7d29310467f146/templates/tfengine/recipes/deployment.hcl#L31) is set to `false` for `terraform_addons` object.

xingao267:
@ernestognw this is the issue I mentioned in our meeting.

PR #986, opened 2021-08-03
Disallow additionalProperties for terraform_addons (#951)

newtonnewton:


MartinPetkov:
Can you run `./scripts/regen.sh` and confirm that there are no errors raised?

newtonnewton:
> Can you run `./scripts/regen.sh` and confirm that there are no errors raised?

just ran this and it prompted me this: 
- (root).terraform_addons: Additional property providers is not allowed
- (root).terraform_addons: Additional property raw_config is not allowed

newtonnewton:
> > Can you run `./scripts/regen.sh` and confirm that there are no errors raised?
> 
> just ran this and it prompted me this:
> 
> * (root).terraform_addons: Additional property providers is not allowed
> * (root).terraform_addons: Additional property raw_config is not allowed

probably because I need to add those to the project recipe as it now supersedes the depolyment recipe it references

xingao267:
That's something I was worried about as well. By setting allow additional properties to `false` in the `project.hcl`, it will not allow any attributes to be specified in the `terraform_addon` block. And I don't think we want to duplicate the same schema for `terraform_addon` in `project.hcl`, it should just use the schema specified in the `deployment.hcl`. `project.hcl` references `deployment.hcl`.

MartinPetkov:
I suspect the errors are just because no properties have been configured for terraform_addons. We should probably define it as an object and define properties it can accept.

MartinPetkov:
I guess the reason we're seeing this behaviour is because project.hcl lets it be anything, but if you put an extra property, it doesn't get passed directly to deployment.hcl.

Could we try that, passing terraform_addons directly to deployment.hcl in line 158?

MartinPetkov:
Another solution I can think of is listing the properties but letting them be anything and letting them be validated in deployment.hcl

MartinPetkov:
Does this fix the issue?

newtonnewton:
> Does this fix the issue?

yes It now catches exceptional attributes as expected and let pass those specified

templates/tfengine/recipes/project.hcl
```diff
@@ -149,6 +149,15 @@ schema = {
         Additional Terraform configuration for the project deployment.
         For schema see ./deployment.hcl.
       EOF
+      type = "object"
+      properties = {
+        raw_config = {}
+	providers = {}
```
MartinPetkov, line 155: Could you make the indentation here consistent?

templates/tfengine/recipes/project.hcl
```diff
@@ -149,6 +149,15 @@ schema = {
         Additional Terraform configuration for the project deployment.
         For schema see ./deployment.hcl.
       EOF
+      type = "object"
+      properties = {
```
MartinPetkov, line 153: Does this also work if you use patternProperties?

MartinPetkov: I'm imagining:
```
patternProperties = {
  "^raw_config|providers|...$" = {}
}
```

MartinPetkov (approved):

xingao267:
Does add `additionalProperties = false` [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/deployment.hcl#L18) do anything?

MartinPetkov:
> Does add `additionalProperties = false` [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/deployment.hcl#L18) do anything?

I tried it, and no, it doesn't help because the properties in project.hcl aren't directly passed to deployment.hcl, they are inherited.

newtonnewton:
> Does add `additionalProperties = false` [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/deployment.hcl#L18) do anything?

just tried, it doesn't work with that unfortunately

xingao267 (approved):

xingao267:
> > Does add `additionalProperties = false` [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/deployment.hcl#L18) do anything?
> 
> I tried it, and no, it doesn't help because the properties in project.hcl aren't directly passed to deployment.hcl, they are inherited.

Hmm ok. But these checks for [inner attributes](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/deployment.hcl#L45) work as expected?

newtonnewton:
> > > Does add `additionalProperties = false` [here](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/deployment.hcl#L18) do anything?
> > 
> > 
> > I tried it, and no, it doesn't help because the properties in project.hcl aren't directly passed to deployment.hcl, they are inherited.
> 
> Hmm ok. But these checks for [inner attributes](https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/deployment.hcl#L45) work as expected?

Just tried that and they don't work as expected, the enforcement is effective on the immediate properties of terraform_addons but not their children

MartinPetkov:
In that case we should see if the fields can be passed directly to deployment.hcl i.e. with a data block.

newtonnewton:
it should all be working as expected now after wrapping all `terraform_addons` into `deployment` and loading flattened 'depolyment' field into its recipe

newtonnewton:
> it should all be working as expected now after wrapping all `terraform_addons` into `deployment` and loading flattened 'depolyment' field into its recipe

should've said running ./scripts/regen.sh and manually checking random fields against validations are all working well on my end. Since I don't have access to view the unit-tests, I'm not sure why it's not letting me pass

MartinPetkov:
The unit tests ask that you

```
Please run the following command from the repo root and check in the changes:
./scripts/regen.sh
```

examples/tfengine/modules/team.hcl
```diff
@@ -47,27 +47,29 @@ template "project_secrets" {
...
-          version_constraints = "~> 3.0"
-        }
-      ]
-      raw_config = <<EOF
-resource "random_password" "db" {
-  length = 16
-  special = true
-}
-EOF
+    deployment = {
```
xingao267, line 50: This is breaking change for all users. I don't think we should require everyone to change their config in order to fix this bug. The fix should be silently in our logic without user-facing changes.

newtonnewton: In that case, it seems we can hardcode this in dumpTemplate. i.e. when ti.Name = `project` do sth...

any thoughts?

xingao267: Did you figure out why resources recipe works but not deployment recipe? We didn't need to hardcode `ti.Name=project ...` in order to make that work.

newtonnewton: the reason is essentially this:

1. we need to pass the flattened data by key `terraform_addons` (almost) so that `ti.Data' wouldn't be empty for the validation (when I showed you guys yesterday I turned on the printing statement at another place which was not what I meant to...)
2. but passing by `terraform_addons` has a problem that it passes all the value `terraform_addons` holds as a key in `depolyment`, namely we lose the `terraform_addons` itself as the key of the map in the validation layer so that the validation process doesn't recognize it. If you compare to `resources`, we pass by the key `resources` so that all the actual key of the `resources` get passed down to validation as opposed to (`providers`, `raw_config` etc subfields of `terraform_addons`) -- sorry a little bit clunky hopefully it makes sense
3. the most straightforward way to salvage this is wrapping all the `terraform_addons` in `deployment` and pass by key `deployment` so on the next level it would have the desired key `terraform_addons` to validate against. 
4. But since you pointed out we would't want to entail such breaking change we may have to hardcode it in `dumpTemplate`. Something I can think of is when `ti.Name == "project"`, create another data carrier `d` and  make `d["terraform_addons"]` carry all the scattered fields in `ti.Data` that used to belong to `terraform_addons`. I know this is grotesque though

MartinPetkov: I'm not sure I understand what the concern is. It seems to me that if we add a new function (passthrough) and only use it here in this one place (project.hcl) in order to pass terraform_addons to deployment.hcl, then it won't affect any other recipes.

xingao267: I played with it a little bit. See https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/991.

newtonnewton: > I'm not sure I understand what the concern is. It seems to me that if we add a new function (passthrough) and only use it here in this one place (project.hcl) in order to pass terraform_addons to deployment.hcl, then it won't affect any other recipes.

I got wrong what you meant. yeah that should work too

xingao267: passthrough sgtm as well.

MartinPetkov:
FYI I think we're proceeding with https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/1003 instead.

PR #980, opened 2021-08-02
Apply BP3 spec to project recipe

ernestognw:
Closes #914

PR #989, opened 2021-08-10
Fix regex for remote state data change

MartinPetkov:


xingao267 (approved):

internal/tfengine/tfengine.go
```diff
@@ -250,7 +250,7 @@ var backendRE = regexp.MustCompile(`(?sU)backend "gcs" {.*}`)
 var remoteStateResourceRE = regexp.MustCompile(`(?sU)data "terraform_remote_state" .* {.*}\n}`)
 
 // remoteStateDataRE is a regex to capture usages of terraform_remote_state in configs.
-var remoteStateDataRE = regexp.MustCompile(`data.terraform_remote_state.*`)
+var remoteStateDataRE = regexp.MustCompile(`\${data\.terraform_remote_state\..*}`)
```
ernestognw, line 253: Actually the `${}` was one of my tries to make it work, but I think this might be:

```suggestion
var remoteStateDataRE = regexp.MustCompile(`data\.terraform_remote_state\..*`)
```

MartinPetkov: If you don't have the ${} in the regex, you end up with `${000}` which is not valid.

MartinPetkov: The issue you were running into is that `.*` also catches the `:` and "if false" value of the ternary operator.

ernestognw: Got it. That's happening on master, but what I'm trying might require changes:

https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/980/commits/31ac2414064222c53a99eb9cb49b82dda58795ef#diff-4c920ffb0b134a0466c5ebd1796cecfa16a2050cf0f1eb91673ef94aa1402c52R55

I'm rendering the variable without the `${}` so it can also match with the `isDotNotation` regex

I'll approve but I guess I might have to reimplement this on my branch

ernestognw (approved):

internal/tfengine/tfengine.go
MartinPetkov: You're right, this isn't complete yet. I'm iterating locally and should have the final one pushed up soon.

MartinPetkov: PTAL

PR #990, opened 2021-08-10
Use go install over go get

MartinPetkov:
Apparently using go get is deprecated for installing packages: https://golang.org/doc/go-get-install-deprecation


MartinPetkov:
I realized this only applies to Go 1.17, and we haven't changed our version to that yet.

PR #991, opened 2021-08-10
Fix schema validation in deployment.hcl recipe.

### LINKED ISSUE #951: deployment.hcl terraform_addons property validation not working

xingao267:
Fixes #951

### DIFF
--- a/templates/tfengine/recipes/deployment.hcl
+++ b/templates/tfengine/recipes/deployment.hcl
@@ -155,7 +155,11 @@ template "terraform" {
   {{end}}
 }
 
-{{if has . "terraform_addons.vars"}}
+// The key "terraform_addons" could be flattened at one level above
+// where "deployment.hcl" recipe is used, e.g. "project.hcl".
+// "deployment.hcl" can also be directly used, in which case "terraform_addons"
+// is not flattened.
+{{if or (has . "vars") (has . "terraform_addons.vars")}}
 template "vars" {
   component_path = "../components/terraform/variables"
   {{if has . "terraform_addons"}}
@@ -166,7 +170,11 @@ template "vars" {
 }
 {{end}}
 
-{{if has . "terraform_addons.outputs"}}
+// The key "terraform_addons" could be flattened at one level above
+// where "deployment.hcl" recipe is used, e.g. "project.hcl".
+// "deployment.hcl" can also be directly used, in which case "terraform_addons"
+// is not flattened.
+{{if or (has . "outputs") (has . "terraform_addons.outputs")}}
 template "outputs" {
   component_path = "../components/terraform/outputs"
   {{if has . "terraform_addons"}}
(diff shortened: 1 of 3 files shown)


templates/tfengine/recipes/deployment.hcl
```diff
@@ -155,7 +155,11 @@ template "terraform" {
   {{end}}
 }
 
-{{if has . "terraform_addons.vars"}}
+// The key "terraform_addons" could be flattened at one level above
+// where "deployment.hcl" recipe is used, e.g. "project.hcl".
+// "deployment.hcl" can also be directly used, in which case "terraform_addons"
+// is not flattened.
+{{if or (has . "vars") (has . "terraform_addons.vars")}}
```
MartinPetkov, line 162: Will this work? If the first case is true (. has "vars"), then there's no "terraform_addons" key to flatten, so does it just get passed down directly?

newtonnewton:
I just tried to run this on my end, It doesn't seem to enforce the validation as expected. 
If you try adding an extra field to items of providers. `{
          name = "google",
          version_constraints = ">=3.0, <= 3.71",
          fakedata = "fakedata"
        }`
it's not enforcing the `additionalProperties = false` on `deployment` level

templates/tfengine/recipes/deployment.hcl
xingao267: Yes. Note this is in `deployment.hcl` and values here will be passed down to each components (e.g. variables, states, etc).

There are two cases where this part will be reached:

1. `deployment.hcl` is called from `project.hcl`, and `terraform_addons` key is flattened/removed there. In that case, the first case in this `or` statement will be hit, the proper component will still be added, and the additional `flatten` statement below will not be hit as the if condition there will be false.

2. `deployment.hcl` is called directly, e.g. in `modules/team.hcl` in the kubernetes template. In that case, second case in this `or` statement will be hit, and the `flatten` operation below will take effect.

xingao267:
> I just tried to run this on my end, It doesn't seem to enforce the validation as expected.
> If you try adding an extra field to items of providers. `{ name = "google", version_constraints = ">=3.0, <= 3.71", fakedata = "fakedata" }`
> it's not enforcing the `additionalProperties = false` on `deployment` level

hmm right. Ok. I got confused about checking for `providers`. That's working because of adding the additionalProperty=false in project.hcl. The inner attributes still not working.

PR #997, opened 2021-08-12
Remove project recipe reference in audit recipe

ernestognw:
## Description

Removed the reference that `audit` recipe is doing to the `project` recipe. This will avoid future duplication of variables at `variables.tf` and `terraform.tfvars`


MartinPetkov:
The description says it removes references to the project recipe, but in the first file I see 

```
name = var.project.project_id
```

Am I misunderstanding the description?

MartinPetkov (approved):
LGTM just have to fix test failures

PR #1003, opened 2021-08-13
Add Passthrough support

MartinPetkov:
This allows directly passing keys from one recipe to the next.

https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/951

### DIFF
--- a/internal/tfengine/schema.go
+++ b/internal/tfengine/schema.go
@@ -113,6 +113,21 @@ properties = {
             }
           }
         }
+
+        passthrough = {
+          description = <<EOF
+            Keys to pass directly to a child template data.
+            This is not recursive.
+
+            Example:
+              The deployment recipe expects "terraform_addons" to be passed to
+	      it directly, so the project recipe uses this key to do so.
+          EOF
+          type = "array"
+          items = {
+            type = "string"
+          }
+        }
       }
     }
   }
--- a/internal/tfengine/schema.go
+++ b/internal/tfengine/schema.go
@@ -121,7 +121,7 @@ properties = {
 
             Example:
               The deployment recipe expects "terraform_addons" to be passed to
-	      it directly, so the project recipe uses this key to do so.
+              it directly, so the project recipe uses this key to do so.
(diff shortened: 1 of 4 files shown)


internal/tfengine/tfengine.go
```diff
@@ -178,6 +179,14 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
 		return err
 	}
 
+	// Pass through specified keys that should be validated against the
+	// schema.
+	for _, k := range ti.Passthrough {
+		if v, ok := conf.Data[k]; ok {
```
xingao267, line 185: should we throw an error if the specified key in passthrough is invalid?

internal/tfengine/schema.go
```diff
@@ -113,6 +113,21 @@ properties = {
...
         }
+
+        passthrough = {
+          description = <<EOF
+            Keys to pass directly to a child template data.
+            This is not recursive.
+
+            Example:
+              The deployment recipe expects "terraform_addons" to be passed to
+	      it directly, so the project recipe uses this key to do so.
```
xingao267, line 124: fix indentation

internal/tfengine/tfengine.go
MartinPetkov: Yes, that's probably a good idea. If it's being passthrough'd, then it's probably necessary.

internal/tfengine/tfengine.go
```diff
@@ -178,6 +179,16 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
...
+			ti.Data[k] = v
+		} else {
+			return fmt.Errorf("did not find key %q to pass through to child template", k)
```
xingao267, line 188: we can probably do 

```
v, ok := conf.Data[k]
if !ok {
   return ...
}
ti.Data[k] = v
```
to reduce indentation

xingao267 (approved):

PR #1004, opened 2021-08-16
Add passthrough to the project recipe

MartinPetkov:
Also removed the error checking, it doesn't work here because one of the fields is optional and isn't always provided. It might be simpler this way and to just let the schema in the child recipe set certain fields as required or not.

Fixes https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/issues/951

### DIFF
--- a/internal/tfengine/tfengine.go
+++ b/internal/tfengine/tfengine.go
@@ -182,11 +182,9 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
 	// Pass through specified keys that should be validated against the
 	// schema.
 	for _, k := range ti.Passthrough {
-		v, ok := conf.Data[k]
-		if !ok {
-			return fmt.Errorf("did not find key %q to pass through to child template", k)
+		if v, ok := conf.Data[k]; ok {
+			ti.Data[k] = v
 		}
-		ti.Data[k] = v
 	}
 
 	switch {
--- a/internal/tfengine/tfengine.go
+++ b/internal/tfengine/tfengine.go
@@ -179,8 +179,9 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
 		return err
 	}
 
-	// Pass through specified keys that should be validated against the
-	// schema.
+	// Pass through keys that should be validated against the schema.
+	// Don't do error checking if keys are missing, instead let schemas
+	// in the child templates do the validation.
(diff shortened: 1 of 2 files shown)


xingao267:
To fully fix 951, should we list out top level attributes e.g. providers in the terraform_addons block in the project.hcl?

internal/tfengine/tfengine.go
```diff
@@ -182,11 +182,9 @@ func dumpTemplate(conf *Config, pwd, cacheDir, outputPath string, ti *templateIn
 	// Pass through specified keys that should be validated against the
 	// schema.
 	for _, k := range ti.Passthrough {
-		v, ok := conf.Data[k]
-		if !ok {
-			return fmt.Errorf("did not find key %q to pass through to child template", k)
+		if v, ok := conf.Data[k]; ok {
```
xingao267, line 185: merging issue?

MartinPetkov: No, see the description. It doesn't work with error checking because the "state_path_prefix field" isn't always passed in. I was thinking it might be simpler to just let schemas in children be responsible for throwing errors if fields aren't passed.

xingao267: I see. Can you add a comment for that in the code as well?

MartinPetkov:
> To fully fix 951, should we list out top level attributes e.g. providers in the terraform_addons block in the project.hcl?

No, I don't think so. I tested by naming "providers" and one of its subfields the wrong way, and it still failed during the schema validation check.

xingao267:
> > To fully fix 951, should we list out top level attributes e.g. providers in the terraform_addons block in the project.hcl?
> 
> No, I don't think so. I tested by naming "providers" and one of its subfields the wrong way, and it still failed during the schema validation check.

Does it complain both issues in the error output?

xingao267 (approved):

MartinPetkov:
> > > To fully fix 951, should we list out top level attributes e.g. providers in the terraform_addons block in the project.hcl?
> > 
> > 
> > No, I don't think so. I tested by naming "providers" and one of its subfields the wrong way, and it still failed during the schema validation check.
> 
> Does it complain both issues in the error output?

No, and I don't think it would. If the schema says "providers" should have a key "name", and instead you gave it "foo" with key "bar", it wouldn't tell you that "provider" should not have key "name" since you didn't give key "provider", it would just tell you "Additional property foo is not allowed"

PR #1005, opened 2021-08-17
Apply terraformPattern function

### LINKED ISSUE #987: Change pattern for parent_id at project.hcl
As seen in: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/980#discussion_r682706191, the `parent_id` property should use a pattern.

The issue is that `mutli_envs.hcl` example uses an output (data.terraform_remote_state.folders.outputs.folder_ids[...]) that doesn't match the pattern.

To solve this issue it is needed to add a way to detect when there is an output used, or replacing the multi_env to use a variable instead.

ernestognw:
## Description

Implemented `terraformPattern` function to render the correct pattern in the template. The objective is to avoid maintenance of specific fields and make the templates smart enough to render the pattern depending on what's available.

Closes #987

### DIFF
--- a/internal/template/funcmap.go
+++ b/internal/template/funcmap.go
@@ -177,7 +177,7 @@ func isDotNotation(s string) (bool, error) {
 // Schema patterns may differ from the terraform pattern for various reasons such as:
 // - Terraform pattern might accept empty strings for optional fields
 // - Terraform pattern might have escaped characters that don't work when rendered in the variables.tf
-// It is recommended to use this function for pattern rendering to avoid redundant maintainance
+// It is recommended to use this function for pattern rendering to avoid redundant maintainance.
 func terraformPattern(m map[string]interface{}) interface{} {
 	return get(m, "terraformPattern", get(m, "pattern"))
 }
--- a/internal/template/funcmap.go
+++ b/internal/template/funcmap.go
@@ -37,6 +37,7 @@ var funcMap = map[string]interface{}{
 	"makeSlice":         makeSlice,
 	"schemaDescription": schemaDescription,
 	"isDotNotation":     isDotNotation,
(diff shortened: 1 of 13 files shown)


MartinPetkov (approved):

internal/template/funcmap.go
```diff
@@ -170,3 +171,9 @@ func isDotNotation(s string) (bool, error) {
 	// - data["foo"][123].bar["baz"].foo
 	return regexp.MatchString(`^([\w]+(\[(("\w+")|\d+)\])*\.[\w]+(\[(("\w+")|\d+)\])*)+$`, s)
 }
+
+// This function returns either "terraformPattern" if present, or the
```
xingao267, line 175: can you add some more documents around why is it needed? i.e. the validation pattern for tfengien config might be different from the validation pattern for terraform configs. Something like that.

ernestognw: Done!

internal/template/funcmap.go
```diff
@@ -173,7 +173,11 @@ func isDotNotation(s string) (bool, error) {
 }
 
 // This function returns either "terraformPattern" if present, or the
-// regular "pattern" otherwise
+// regular "pattern" otherwise.
+// Schema patterns may differ from the terraform pattern for various reasons such as:
+// - Terraform pattern might accept empty strings for optional fields
+// - Terraform pattern might have escaped characters that don't work when rendered in the variables.tf
+// It is recommended to use this function for pattern rendering to avoid redundant maintainance
```
xingao267, line 180: nit: period at the end of sentence

xingao267 (approved):

internal/template/funcmap.go
ernestognw: Added 🤘

PR #1006, opened 2021-08-17
Simplify variables and rendering for CICD

### LINKED ISSUE #970: Recalculate terraform_root_prefix for cicd recipe
On the CICD recipe, there are two variables too similar:

`terraform_root`
`terraform_root_prefix`

These variables are passed to the trigger submodules used on the recipe and may be simplified.

Started at #949

ernestognw:
## Description

Revisited CICD recipe to apply schema variables completely and simplify extra conditional renderings.

Closes #970 and #958 

Note: I decided to send this PR before completely addressing #956 to make both easier to review


templates/tfengine/components/cicd/triggers.tf
```diff
@@ -1,327 +0,0 @@
-# Copyright 2021 Google LLC
```
ernestognw, line 1: I think this one made it into the repo again after some merge I did. I'm removing it again but It wasn't supposed to be here in the first place

templates/tfengine/components/cicd/variables.tf
```diff
@@ -12,6 +12,28 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    condition     = can(regex("{{$props.parent_id.pattern}}", var.parent_id))
+    error_message = "The parent_id must be valid. Should have only numeric values with a length between 8 and 25 digits. See https://cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy to know how to get your organization/folder id."
+  }
+}
+
+variable "parent_type" {
+  type        = string
+  description = {{schemaDescription $props.parent_type.description}}
+  validation {
+    condition     = can(regex("{{$props.parent_type.pattern}}", var.parent_type))
```
MartinPetkov, line 32: Should use the terraformPattern function here and in parent_id

templates/tfengine/components/cicd/variables.tf
```diff
@@ -78,7 +112,34 @@ variable "envs" {
...
+    *** run_on_push = {{$triggerProps.apply.properties.run_on_push.description}}
+    *** run_on_schedule = {{$triggerProps.apply.properties.run_on_schedule.description}}
+    ** plan = {{$triggerProps.plan.description}}
+    *** skip = Whether or not to skip creating trigger resources.
+    *** run_on_push = {{$triggerProps.plan.properties.run_on_push.description}}
+    *** run_on_schedule = {{$triggerProps.plan.properties.run_on_schedule.description}}
+    ** validate = {{$triggerProps.validate.description}}
+    *** skip = Whether or not to skip creating trigger resources.
+    *** run_on_push = {{$triggerProps.validate.properties.run_on_push.description}}
+    *** run_on_schedule = {{$triggerProps.validate.properties.run_on_schedule.description}}
```
MartinPetkov, line 135: I think spaces (2 per level) would look better than multiple asterisks

MartinPetkov (approved):

templates/tfengine/components/cicd/variables.tf
ernestognw: You're right. I'm sorry. Started with this one first and forgot it 😅

templates/tfengine/components/cicd/variables.tf
ernestognw: You're right too, also that's what we agreed for other recipes.

Fixed

templates/tfengine/components/cicd/main.tf
```diff
@@ -117,12 +115,11 @@ resource "google_project_iam_member" "cloudbuild_builds_viewers" {
     google_project_service.services,
   ]
 }
-{{- end}}
-
-{{- if has . "build_editors"}}
 
 # IAM permissions to allow approvers to edit/create the cloud build jobs.
 resource "google_project_iam_member" "cloudbuild_builds_editors" {
+  count    = length(var.build_editors) > 0 ? 1 : 0
```
xingao267, line 121: This is not needed i believe. This resource already has a for_each block. Same thing for the build viewers.

templates/tfengine/components/cicd/main.tf
```diff
@@ -214,55 +210,57 @@ resource "google_service_account" "cloudbuild_scheduler_sa" {
...
-  for_each = toset(local.cloudbuild_sa_viewer_roles)
-  {{- end}}
-  {{- if eq (get . "parent_type") "organization"}}
-  org_id   = {{.parent_id}}
-  {{- else}}
-  folder   = {{.parent_id}}
-  {{- end}}
+# Grant Cloud Build Service Account access to the organization.
+resource "google_organization_iam_member" "cloudbuild_sa_organization_iam" {
+  for_each = var.parent_type == "organization" ? local.has_apply_jobs ? toset(local.cloudbuild_sa_editor_roles) : toset(local.cloudbuild_sa_viewer_roles) : []
```
xingao267, line 249: Add () around `local.has_apply_jobs ? toset(local.cloudbuild_sa_editor_roles) : toset(local.cloudbuild_sa_viewer_roles)` so it's more clear that logic happens first?

templates/tfengine/components/cicd/main.tf
```diff
@@ -214,55 +210,57 @@ resource "google_service_account" "cloudbuild_scheduler_sa" {
...
+  role     = each.value
+  member   = local.cloudbuild_sa
+  depends_on = [
+    google_project_service.services,
+  ]
+}
+
+# Grant Cloud Build Service Account access to the folder.
+resource "google_folder_iam_member" "cloudbuild_sa_folder_iam" {
+  for_each = var.parent_type == "folder" ? local.has_apply_jobs ? toset(local.cloudbuild_sa_editor_roles) : toset(local.cloudbuild_sa_viewer_roles) : []
```
xingao267, line 261: same here

templates/tfengine/components/cicd/terraform.tfvars
```diff
@@ -98,5 +96,6 @@ envs = [
         }
     }
   },
-  {{- end}}
+  {{- end}}
```
xingao267, line 99: trailing space

templates/tfengine/components/cicd/terraform.tfvars
```diff
@@ -98,5 +96,6 @@ envs = [
         }
     }
   },
-  {{- end}}
+  {{- end}} 
 ]
+{{hclField . "grant_automation_billing_user_role" -}}
```
xingao267, line 101: Move this up. Putting this at the end especially below the big github and csr block will make it easily ignored.

templates/tfengine/recipes/cicd.hcl
```diff
@@ -21,6 +21,25 @@ schema = {
     "scheduler_region",
   ]
   properties = {
+    billing_account = {
+      description = "ID of billing account to attach to this project."
+      type        = "string"
+    }
```
xingao267, line 27: why is this added now?

templates/tfengine/components/cicd/triggers/apply/main.tf
```diff
@@ -11,6 +11,11 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License. */ -}}
+locals {
+  terraform_root = var.terraform_root == "/" ? "." : var.terraform_root
+  terraform_root_prefix = var.terraform_root == "/" ? "" : "${var.terraform_root}/"
```
xingao267, line 16: I'm looking at https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/components/cicd/triggers.tf#L21 and i think there is subtle differences there. You probably should use local.terraform_root here so the case where `var.terraform_root` (or `local.terraform_root`) is `.` is considered. And in that case, `terraform_root_prefix` is set to `""`.

templates/tfengine/components/cicd/main.tf
ernestognw: You're right, I'm also questioning about what will happen if both are specified 🧐

Fixed anyway

templates/tfengine/components/cicd/main.tf
ernestognw: Done

I also tried to separate it into many lines but it doesn't work. The `terraform validate` throws an error.
If you know any possible solution to make it more readable let me know. I'm resolving this thread.

templates/tfengine/components/cicd/terraform.tfvars
ernestognw: Done

templates/tfengine/recipes/cicd.hcl
ernestognw: It wasn't there but it wasn't needed to afaik.

I added it because I needed the `description` field.

I think it wasn't added because it is not really required, but is provided to the higher order `data = {}` block at the examples

templates/tfengine/components/cicd/triggers/apply/main.tf
ernestognw: You're right, I changed it for every trigger module.

Thanks for checking it out with that detail. I'm resolving

templates/tfengine/components/cicd/triggers/validate/main.tf
```diff
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License. */ -}}
 locals {
   terraform_root = var.terraform_root == "/" ? "." : var.terraform_root
-  terraform_root_prefix = var.terraform_root == "/" ? "" : "${var.terraform_root}/"
+  terraform_root_prefix = local.terraform_root == "." ? "" : "${var.terraform_root}/"
```
xingao267, line 16: can we also use local.terraform in the "${local.terraform_root}/" as well? I think logically it won't make a difference, but more consistent if we use one terraform_root in terraform_root_prefix value definition.

ernestognw: Sure, looks better to me too.

Done

xingao267 (approved):

PR #1008, opened 2021-08-18
Allow specifying force_destroy on storage_bucket resource

### LINKED ISSUE #973: Allow specifying force_destroy to true
allow specifying https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket#force_destroy to true for a storage bucket.

ernestognw:
## Description

Added force_destroy option to storage_buckets resource schema

Closes #973

### DIFF
--- a/examples/tfengine/resources_only.hcl
+++ b/examples/tfengine/resources_only.hcl
@@ -97,6 +97,7 @@ template "resources" {
       }]
       storage_buckets = [{
         name = "example-prod-bucket"
+        force_destroy = true
         labels = {
           type = "phi"
         }
--- a/examples/tfengine/resources_only.hcl
+++ b/examples/tfengine/resources_only.hcl
@@ -97,7 +97,7 @@ template "resources" {
       }]
       storage_buckets = [{
         name = "example-prod-bucket"
-        force_destroy = true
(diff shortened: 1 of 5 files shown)


examples/tfengine/resources_only.hcl
```diff
@@ -97,6 +97,7 @@ template "resources" {
       }]
       storage_buckets = [{
         name = "example-prod-bucket"
+        force_destroy = true
```
xingao267, line 100: can you add a comment above saying it's for example purpose only. Generally, it's not recommended as it would delete objects in the bucket if any...

ernestognw: I think I can set it to false and add the # For example purposes only

It won't really affect since the default is false but it would be rendered during tests, which I think is the purpose of specifying it in at least one example.

Wdyt?

MartinPetkov (approved):

examples/tfengine/resources_only.hcl
xingao267: sgtm

ernestognw: Done

xingao267 (approved):

PR #1009, opened 2021-08-18
Merge triggers for CICD recipe

### LINKED ISSUE #956: Avoid repeated resources on cicd recipe

ernestognw:
# Description

Attempt to merge triggers into a single submodule to avoid repeated code.

Closes #956


templates/tfengine/components/cicd/main.tf
```diff
@@ -43,6 +43,8 @@ locals {
   cloudbuild_sa = "serviceAccount:${data.google_project.devops.number}@cloudbuild.gserviceaccount.com"
   has_scheduled_jobs = anytrue([for env in var.envs : env.triggers.validate.run_on_schedule || env.triggers.plan.run_on_schedule || env.triggers.apply.run_on_schedule])
   has_apply_jobs = anytrue([for env in var.envs : !env.triggers.apply.skip])
+  is_github = var.github.name != ""
+  is_cloud_source_repository = !var.is_github && var.cloud_source_repository.name != ""
```
ernestognw, line 47: I'm doing this to ensure that github and csr are mutually exclusive since now both are specified on `variables.tf`

There's no way to enforce that rather than doing something like this, and I put preference over github because csr is the one that adds more extra resources.

templates/tfengine/components/cicd/envs/main.tf
```diff
@@ -12,64 +12,79 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
   terraform_root          = var.terraform_root
   service_account_email   = var.service_account_email
+
+  push = {
+    skip = true # Always true since scheduled trigger may act as a push trigger too
+    disabled = true
+  }
+
+  scheduled = {
+    skip = var.apply.validate.skip
```
MartinPetkov, line 87: Should this be var.triggers.apply?

ernestognw: Thanks! That was a very obvious one and still didn't noticed

templates/tfengine/components/cicd/main.tf
```diff
@@ -265,31 +260,30 @@ resource "google_folder_iam_member" "cloudbuild_sa_folder_iam" {
 }
 
 # Create Google Cloud Build triggers for specified environments
-module "triggers" {
+module "environment_triggers" {
   for_each = {for env in var.envs : env.name => env}
-  // TODO(ernestognw): Merge triggers to simplify resources #956
-  source  = "./triggers"
+  source  = "./envs"
```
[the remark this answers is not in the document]
ernestognw: Agree, thanks!

MartinPetkov:
Looking at the final result, I'm inclined to think it's better to just keep templating and have clearer final output files. I don't like that when I look at a triggers/main.tf file, I see
```
github = var.is_github ? ... : []
trigger_template = is_cloud_source_repository ? ... : []
```
and similar things elsewhere. It makes me have to mentally do the boolean logic by looking at the terraform.tfvars to figure out what kind of trigger this is and what it's doing.

ernestognw:
> Looking at the final result, I'm inclined to think it's better to just keep templating and have clearer final output files. I don't like that when I look at a triggers/main.tf file, I see
> 
> ```
> github = var.is_github ? ... : []
> trigger_template = is_cloud_source_repository ? ... : []
> ```
> 
> and similar things elsewhere. It makes me have to mentally do the boolean logic by looking at the terraform.tfvars to figure out what kind of trigger this is and what it's doing.

On my very first try, this is kind of what I was talking about. Trying to reduce the code creates even more complexity that is more difficult to maintain, but I think it worths the effort to try to accomplish something in the middle.

Particularly with the github/csr logic, that can still be templated while the triggers are still merged? What do you think? 

We can have a meeting with Xin to evaluate. I think these are the possibilities:

1. Keep it like it is right now
2. Keep merged the triggers but templating only the github/csr part
3. Not implementing at all

MartinPetkov:
> > Looking at the final result, I'm inclined to think it's better to just keep templating and have clearer final output files. I don't like that when I look at a triggers/main.tf file, I see
> > ```
> > github = var.is_github ? ... : []
> > trigger_template = is_cloud_source_repository ? ... : []
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and similar things elsewhere. It makes me have to mentally do the boolean logic by looking at the terraform.tfvars to figure out what kind of trigger this is and what it's doing.
> 
> On my very first try, this is kind of what I was talking about. Trying to reduce the code creates even more complexity that is more difficult to maintain, but I think it worths the effort to try to accomplish something in the middle.
> 
> Particularly with the github/csr logic, that can still be templated while the triggers are still merged? What do you think?
> 
> We can have a meeting with Xin to evaluate. I think these are the possibilities:
> 
> 1. Keep it like it is right now
> 2. Keep merged the triggers but templating only the github/csr part
> 3. Not merging at all

A meeting is a good idea, but sure, #2 sounds like it could be an improvement.

ernestognw:
> > > Looking at the final result, I'm inclined to think it's better to just keep templating and have clearer final output files. I don't like that when I look at a triggers/main.tf file, I see
> > > ```
> > > github = var.is_github ? ... : []
> > > trigger_template = is_cloud_source_repository ? ... : []
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and similar things elsewhere. It makes me have to mentally do the boolean logic by looking at the terraform.tfvars to figure out what kind of trigger this is and what it's doing.
> > 
> > 
> > On my very first try, this is kind of what I was talking about. Trying to reduce the code creates even more complexity that is more difficult to maintain, but I think it worths the effort to try to accomplish something in the middle.
> > Particularly with the github/csr logic, that can still be templated while the triggers are still merged? What do you think?
> > We can have a meeting with Xin to evaluate. I think these are the possibilities:
> > 
> > 1. Keep it like it is right now
> > 2. Keep merged the triggers but templating only the github/csr part
> > 3. Not merging at all
> 
> A meeting is a good idea, but sure, #2 sounds like it could be an improvement.

I'll open a separate PR against this branch with the #2 so we can compare the alternatives before setting up the meeting.

I'll let you know!

templates/tfengine/components/cicd/main.tf
xingao267: Add a comment saying these two are mutually exclusive

templates/tfengine/components/cicd/main.tf
```diff
@@ -265,21 +260,18 @@ resource "google_folder_iam_member" "cloudbuild_sa_folder_iam" {
 }
 
 # Create Google Cloud Build triggers for specified environments
-module "triggers" {
+module "environment_triggers" {
```
xingao267, line 263: module "triggers" should be enough

templates/tfengine/components/cicd/env/main.tf
```diff
@@ -12,64 +12,79 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
```
xingao267, line 12: this file can be called cicd/triggers/main.tf

and your current triggers/main.tf can be renamed to trigger/main.tf

ernestognw: Are you sure? The `triggers/main.tf` actually has more than 1 trigger.

Is just a semantic problem, not a big deal to leave it as you request, but I want to make you aware

xingao267: discussed offline, we can keep as is.

templates/tfengine/components/cicd/env/main.tf
```diff
@@ -12,64 +12,79 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-  {{- if has . "github"}}
   github                  = var.github
-  {{- end}}
   project_id              = var.project_id
   scheduler_region        = var.scheduler_region
   terraform_root          = var.terraform_root
   service_account_email   = var.service_account_email
+
+  push = {
+    skip = true # Always true since scheduled trigger may act as a push trigger too
```
xingao267, line 82: # Always true since the scheduled trigger act as the push trigger for "apply".

templates/tfengine/components/cicd/env/triggers/main.tf
```diff
@@ -0,0 +1,120 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. */ -}}
+locals {
+  terraform_root = var.terraform_root == "/" ? "." : var.terraform_root
+  terraform_root_prefix = local.terraform_root == "." ? "" : "${local.terraform_root}/"
+  // Github and CSR are mutually exclusive so there shouldn't be both specified on the same
+  // terraform configuration, but in case they're, priority goes to github since it creates the
+  // less amount of extra resources
```
xingao267, line 19: Creates the less amount of extra resources don't seem to me a very convincing reason. You can just mention the priority goes to GitHub. That's good enough.

ernestognw: Sg, done

xingao267 (approved):

MartinPetkov (approved):

templates/tfengine/recipes/cicd.hcl
```diff
@@ -58,10 +64,12 @@ schema = {
         owner = {
           description = "GitHub repo owner."
           type        = "string"
+          default     = "\"\""
```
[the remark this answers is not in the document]
xingao267: this one is missed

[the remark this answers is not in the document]
ernestognw: Added to all of the same defaults

templates/tfengine/components/cicd/env/triggers/main.tf
```diff
@@ -80,18 +79,18 @@ resource "google_cloudbuild_trigger" "plan_scheduled" {
...
   substitutions = {
     _TERRAFORM_ROOT = local.terraform_root
     _MANAGED_DIRS = var.managed_dirs
   }
 }
 
-resource "google_cloud_scheduler_job" "plan_scheduler" {
-  count     = (!var.skip && var.run_on_schedule != "") ? 1 : 0
+resource "google_cloud_scheduler_job" "scheduler" {
+  count     = (var.scheduled.skip || var.scheduled.disabled) ? 0 : 1
```
xingao267, line 91: will this logic cause issues?

For validate and plan triggers, var.scheduled.disabled is always set to true.

PR #1011, opened 2021-08-24
Add more logging to the deletion check

MartinPetkov:
Sometimes this check fails with "parse error: Invalid numeric literal at line 1, column 2". This looks like it's coming from jq failing to parse the output of terraform show, so log that output separately for debugging.


xingao267 (approved):
at some point we probably should add a flag to this script to control whether or not to print extra debugging info

PR #1018, opened 2021-08-31
Update tf-deletion-check.sh

jeffbryner:
syntax fix up to avoid the ./cicd/configs/tf-deletion-check.sh: line 77: unexpected EOF while looking for matching `"' error


MartinPetkov (approved):

MartinPetkov:
From the unit tests:
```
Please run the following command from the repo root and check in the changes:
./scripts/regen.sh
```

jeffbryner:
@googlebot I signed it!

jeffbryner:
> From the unit tests:
> 
> ```
> Please run the following command from the repo root and check in the changes:
> ./scripts/regen.sh
> ```

sorry, just realized this script lives in multiple places and checked in what I think is the main file (templates/tfengine/components/cicd/configs). Ran the regen.sh script and there is no diff in the tf-deletion-check.sh in other directories, but I'm not sure the script works

MartinPetkov:
> > From the unit tests:
> > ```
> > Please run the following command from the repo root and check in the changes:
> > ./scripts/regen.sh
> > ```
> 
> sorry, just realized this script lives in multiple places and checked in what I think is the main file (templates/tfengine/components/cicd/configs). Ran the regen.sh script and there is no diff in the tf-deletion-check.sh in other directories, but I'm not sure the script works

The regen script definitely works, but you might need a Go environment.

I've opened https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/1021 instead.

PR #1021, opened 2021-09-01
Fix deletion check syntax and add basic unit test

MartinPetkov:


build/sh_syntax_check.sh
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+# This file ensures that all shell script in the repo have no syntax errors.
+
+#!/usr/bin/env bash
+bad=0
+for script in $(find -name "*.sh"); do
+  bash -n "${script}"
+  if [[ $? != 0 ]]; then
+    bad=1
```
pasha-gh, line 22: Can we move the logic from the next if-statement to here, eliminate the need of using the local variable `bad`, and make the code a bit shorter?

```
  if [[ $? != 0 ]]; then
    echo "Scripts with invalid syntax found, please look at the errors above and fix."
    exit 1
  fi
```

MartinPetkov: No, the idea here is to show all the bad scripts in one go. Otherwise if for some reason multiple scripts are bad, it'll take multiple rounds of  CI/CD runs, commits and lost time. This way you get one lists of bad scripts and can fix them all in one go.

pasha-gh: gotcha

pasha-gh (approved):

PR #1034, opened 2021-09-27
Add support for logs_bucket and service_account to CICD.

pasha-gh:
Please wait for `v3.86.0` terraform provider [release](https://github.com/hashicorp/terraform-provider-google/releases) before merging.

This release will add [support](https://github.com/hashicorp/terraform-provider-google/pull/10159) for `service_account` field in `google_cloudbuild_trigger` resources which is required for this change.


examples/tfengine/modules/foundation.hcl
```diff
@@ -147,6 +181,9 @@ template "cicd" {
           location = "us-east1"
           name     = "cicd-pool"
         }
+        // Do not use $${module.project.project_id} for project ID.
+        service_account = "projects/{{.prefix}}-{{.env}}-devops/serviceAccounts/cloudbuild_sa"
+        logs_bucket = "gs://{{.prefix}}-logs-bucket"
```
xingao267, line 186: `logs_bucket` is mandatory when custom service account is used. We should create the bucket inside the cicd recipe and somehow enforce that dependency. Another way could be always create and use a custom logs bucket. In that case, we should check what's the implication of change in terms of the existing logs bucket.

examples/tfengine/modules/foundation.hcl
```diff
@@ -107,6 +110,37 @@ template "groups" {
...
+        ]
+        "roles/logging.logWriter" = [
+          "serviceAccount:$${google_service_account.cloudbuild_sa.account_id}@{{.prefix}}-{{.env}}-devops.iam.gserviceaccount.com",
+        ]
+      }
+      storage_buckets = [{
+        name = "{{.prefix}}-logs-bucket"
+      }]
+    }
+  }
```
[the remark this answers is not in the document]
pasha-gh: Done. Added requirement conditions in triggers.tf in the CICD component.

examples/tfengine/modules/foundation.hcl
pasha-gh: Done. Added requirement conditions in triggers.tf in the CICD component.

examples/tfengine/modules/foundation.hcl
```diff
@@ -111,6 +114,25 @@ template "cicd" {
   recipe_path = "{{.recipes}}/cicd.hcl"
   output_path = "./cicd"
   data = {
+    resources = {
```
xingao267, line 117: I'm leaning towards hiding these resources inside the cicd recipe and not asking users to write them explicitly here as the purpose of cicd recipe is to expose a simpler interface for users to configure CICD pipelines. Resources here such as the IAM members are implementation details and should be handled automatically by the recipe.

pasha-gh: If we are automatically creating the service account with the required IAM members, then why don't we use the default cloudbuild service account?
The other concern I have is the `logs_bucket` creation. If we don't ask users for any inputs here, we have to pick a name for the bucket. Since the bucket name must be globally unique, this can potentially cause the CICD deployment to fail due to duplicate names.

pasha-gh: Implementation updated.

templates/tfengine/components/cicd/variables.tf
```diff
@@ -25,3 +25,15 @@ variable "state_bucket" {
...
+
+variable "logs_bucket" {
+  description = "Name of the bucket used for cloudbuild logs"
+  default     = "cloudbuild-sa@{{.project_id}}.iam.gserviceaccount.com"
+  type        = string
+}
+
+variable "service_account_email" {
+  description = "Service account used for cloudbuild triggers"
+  default     = "{{.project_id}}-cloudbuild-logs-bucket"
```
xingao267, line 37: The `default` values seem swapped.

templates/tfengine/components/cicd/terraform.tfvars
```diff
@@ -15,3 +15,5 @@
 billing_account = "{{.billing_account}}"
 project_id      = "{{.project_id}}"
 state_bucket    = "{{.state_bucket}}"
+{{hclField . "service_account" -}}
```
xingao267, line 18: The variable name is `service_account_email` but here is `service_account`.

templates/tfengine/components/cicd/main.tf
```diff
@@ -51,7 +51,6 @@ data "google_project" "devops" {
 }
 
 locals {
-  cloudbuild_sa = "serviceAccount:${data.google_project.devops.number}@cloudbuild.gserviceaccount.com"
```
xingao267, line 54: We should still support for the case where users are using the default service account. Also we should make it a non-breaking changes for those users (who use the default service account).

templates/tfengine/components/cicd/main.tf
```diff
@@ -211,7 +210,7 @@ resource "google_project_iam_member" "cloudbuild_sa_project_iam" {
   for_each = toset(local.cloudbuild_devops_roles)
   project  = var.project_id
   role     = each.key
-  member   = local.cloudbuild_sa
+  member   = "serviceAccount:${var.service_account_email}"
```
xingao267, line 213: You can keep the `cloudbuild_sa` local variable, and just gave it different values based on different cases:

1. `service_account_email` is specified, then `local.cloudbuild_sa` will be "serviceAccount:${var.service_account_email}".
2. service_account_email is not specified, then `local.cloudbuild_sa` will be the current value.

templates/tfengine/recipes/cicd.hcl
```diff
@@ -127,6 +127,27 @@ schema = {
       EOF
       type        = "boolean"
     }
+    service_account = {
```
xingao267, line 130: let's use service_account_email here to be consistent

templates/tfengine/components/cicd/main.tf
```diff
@@ -250,14 +249,51 @@ resource "google_project_iam_member" "cloudbuild_scheduler_sa_project_iam" {
 }
 {{- end}}
 
+{{- if not (has . "service_account")}}
+resource "google_service_account" "cloudbuild_sa" {
```
xingao267, line 253: In the case where this service account needs to be created, the other resources that uses this service account should reference it so dependency can be enforced.

templates/tfengine/components/cicd/main.tf
```diff
@@ -250,14 +249,51 @@ resource "google_project_iam_member" "cloudbuild_scheduler_sa_project_iam" {
...
+  project      = var.project_id
+  account_id   = "cloudbuild-sa"
```
xingao267, line 255: Instead of hardcoding the ID here, we can probably just ask users to provide the ID and we always create it in the `devops` project in the cicd recipe. Or another option is instead of asking for a service_account_id, we ask for a boolean value `use_custom_service_account` or something like that. And based on that value, we create or do not create a new SA to be used for cloud build jobs. In that case, we hardcode a value such as `cloudbuild-sa`. The first option is more flexible but second option seems fine to me as well.

Also if users do not provide service_account, then this block of config will be executed, however, in the other places, it's still using `var.service_account_email`. I guess you are relying on the default value of it declared in the `variables.tf`. That's a bit hacky. Also the other resources that use this SA depends on the SA to be created first so reference should be used.

templates/tfengine/components/cicd/variables.tf
xingao267: The default value for the bucket can be invalid if the project ID contains characters that bucket name doesn't like.

https://cloud.google.com/resource-manager/docs/creating-managing-projects
https://cloud.google.com/storage/docs/naming-buckets

Asking the users to provide the bucket name is probably safer. And. we only use the custom logs bucket in the google_cloudbuild_trigger resource when bucket is explicitly asked for and created, otherwise, we can leave that not specified and let cloud build trigger to use the default bucket.

templates/tfengine/components/cicd/main.tf
xingao267: On a second thought, maybe we can enforce the behaviour of always using a custom SA as it's generally a better practice. Let's chat about it a bit more tomorrow.

templates/tfengine/components/cicd/variables.tf
pasha-gh: So we shouldn't create the bucket, right? Either user provides creates one and provides it here, or we use the default bucket?

templates/tfengine/components/cicd/terraform.tfvars
pasha-gh: Fixed.

templates/tfengine/components/cicd/variables.tf
pasha-gh: Resolving based on our conversation.

templates/tfengine/components/cicd/main.tf
pasha-gh: SA and logs_bucket are added to required fields, resolving.

templates/tfengine/components/cicd/main.tf
```diff
@@ -250,6 +254,41 @@ resource "google_project_iam_member" "cloudbuild_scheduler_sa_project_iam" {
...
+}
+
+module "project_iam_members" {
+  source  = "terraform-google-modules/iam/google//modules/projects_iam"
+  version = "~> 7.2.0"
+
+  projects = [var.project_id]
+  mode     = "additive"
+
+  bindings = {
```
xingao267, line 282: These additional bindings should be added to the existing list at the top (in the locals). From looking at https://cloud.google.com/build/docs/securing-builds/configure-user-specified-service-accounts#permissions, `roles/logging.logWriter` is needed to write logs. I think granting it in the `devops` project should be enough, but might need it on the folder level, but we can try on the devops project first. `cloudbuild_devops_roles` is used for this, and we should add the new role there.

Is `roles/iam.serviceAccountUser` needed? The doc says "the user requesting the build" needs this role. I'm not sure why the SA itself need this role, unless i miss anything here.

templates/tfengine/components/cicd/triggers.tf
```diff
@@ -178,6 +186,11 @@ resource "google_cloudbuild_trigger" "plan_{{.name}}" {
   }
   {{- end}}
 
+  // TODO refactor to "projects/${var.project_id}/serviceAccounts/${var.service_account}"
```
xingao267, line 189: remove

docs/tfengine/schemas/cicd.md
```diff
@@ -34,6 +34,9 @@
 | github.name | GitHub repo name. | string | false | - | - |
 | github.owner | GitHub repo owner. | string | false | - | - |
 | grant_automation_billing_user_role | Whether or not to grant automation service account the billing.user role. Default to true. | boolean | false | - | - |
+| logs_bucket | Google Cloud Storage bucket name where logs should be written. The bucket will be created as part of CICD. | string | true | - | - |
```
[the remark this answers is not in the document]
pasha-gh: Done.

docs/tfengine/schemas/cicd.md
```diff
@@ -34,6 +34,9 @@
...
 | project_id | ID of project to deploy CICD in. | string | false | - | ^[a-z][a-z0-9\-]{4,28}[a-z0-9]$ |
 | scheduler_region | [Region](https://cloud.google.com/appengine/docs/locations) where the scheduler job (or the App Engine App behind the sceneces) resides. Must be specified if any triggers are configured to be run on schedule. | string | true | - | - |
+| service_account | The service account ID used for all user-controlled operations. This service account should have several permissions to perform different operations. These permissions will be granted to the service account as part of the CICD deployment. See <https://cloud.google.com/build/docs/securing-builds/configure-user-specified-service-accounts#permissions>. | string | true | - | - |
```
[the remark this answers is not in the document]
pasha-gh: Done.

examples/tfengine/generated/org_foundation/cicd/triggers.tf
```diff
@@ -61,6 +65,11 @@ resource "google_cloudbuild_trigger" "plan_prod" {
     }
   }
 
+  // TODO refactor to "projects/${var.project_id}/serviceAccounts/${var.service_account}"
```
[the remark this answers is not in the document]
pasha-gh: Done.

templates/tfengine/components/cicd/triggers.tf
pasha-gh: Done.

templates/tfengine/components/cicd/main.tf
pasha-gh: Done.
Since we always create the `logs_bucket` in the devops project then granting the permission at the project level should be enough.

templates/tfengine/components/cicd/main.tf
```diff
@@ -250,6 +257,24 @@ resource "google_project_iam_member" "cloudbuild_scheduler_sa_project_iam" {
...
+{{- if not (get .service_account "exists")}}
+resource "google_service_account" "cloudbuild_sa" {
+  project      = var.project_id
+  account_id   = var.service_account
+  display_name = "Cloudbuild service account"
+  description  = "Cloudbuild service account"
+}
+{{- end}}
+
+module "logs_bucket" {
```
xingao267, line 269: Add comments for these resources like the other ones.

templates/tfengine/components/cicd/main.tf
```diff
@@ -250,6 +257,24 @@ resource "google_project_iam_member" "cloudbuild_scheduler_sa_project_iam" {
 }
 {{- end}}
 
+{{- if not (get .service_account "exists")}}
```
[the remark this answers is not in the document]
pasha-gh: Done.

templates/tfengine/components/cicd/triggers.tf
```diff
@@ -109,6 +113,10 @@ resource "google_cloudbuild_trigger" "validate_scheduled_{{.name}}" {
   }
   {{- end}}
 
+  service_account = "projects/${var.project_id}/serviceAccounts/${var.service_account}"
+
+  logs_bucket = "gs://${var.logs_bucket}"
```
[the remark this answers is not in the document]
pasha-gh: Also used referencing.

examples/tfengine/generated/org_foundation/cicd/triggers.tf
```diff
@@ -92,12 +98,15 @@ resource "google_cloudbuild_trigger" "apply_prod" {
...
 
+  service_account = "projects/${var.project_id}/serviceAccounts/${local.cloudbuild_sa_email}"
+
   filename = "terraform/cicd/configs/tf-apply.yaml"
 
   substitutions = {
     _TERRAFORM_ROOT = "terraform"
     _MANAGED_DIRS   = "groups audit example-prod-networks monitor org_policies folders"
     _WORKER_POOL    = ""
+    _LOGS_BUCKET    = "gs://${var.logs_bucket}"
```
xingao267, line 109: Use reference here? so that the race condition will not happen.

templates/tfengine/components/cicd/triggers.tf
```diff
@@ -292,12 +304,15 @@ resource "google_cloudbuild_trigger" "apply_{{.name}}" {
...
 
+  service_account = "projects/${var.project_id}/serviceAccounts/${local.cloudbuild_sa_email}"
+
   filename = "{{$terraform_root_prefix}}cicd/configs/tf-apply.yaml"
 
   substitutions = {
     _TERRAFORM_ROOT = "{{$terraform_root}}"
     _MANAGED_DIRS = "{{$managed_dirs}}"
     _WORKER_POOL = "{{$worker_pool}}"
+    _LOGS_BUCKET = "gs://${var.logs_bucket}"
```
xingao267, line 315: use reference

templates/tfengine/components/cicd/main.tf
```diff
@@ -51,7 +51,11 @@ data "google_project" "devops" {
 }
 
 locals {
-  cloudbuild_sa = "serviceAccount:${data.google_project.devops.number}@cloudbuild.gserviceaccount.com"
+{{- if get .service_account "exists" false}}
```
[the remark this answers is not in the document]
pasha-gh: Added an example in devops.hcl for when the service account exists. However, by default we use, `module.project.project_id` to generate the service account email. Is there a possibility where the SA is not created in the devops project? In that case we should also ask for `service_account_project_id`.

templates/tfengine/components/cicd/main.tf
```diff
@@ -250,14 +257,34 @@ resource "google_project_iam_member" "cloudbuild_scheduler_sa_project_iam" {
...
+
 # Cloud Build - Cloud Build Service Account IAM permissions
 {{- if and $hasApplyJobs (get . "grant_automation_billing_user_role" true)}}
 
 # IAM permissions to allow Cloud Build Service Account use the billing account.
 resource "google_billing_account_iam_member" "binding" {
   billing_account_id = var.billing_account
   role               = "roles/billing.user"
-  member             = local.cloudbuild_sa
+  member             = "serviceAccount:${local.cloudbuild_sa_email}"
```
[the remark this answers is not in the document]
pasha-gh: Since this is also used in `triggers.tf` and there it shouldn't have the `serviceAccount:` part.

PR #1051, opened 2021-11-02
Add storage.objectAdmin role to cloudbuild Service Account.

### LINKED ISSUE #1050: CICD builds failing due to service account lacking log permissions
When running the builds `tf-validate-prod` and `tf-plan-prod` on the CICD pipeline (generated by `folder_foundation.hcl`), the builds fail with the error message:

```Failure setting up GCS logging: failed to create GCS logging client: error creating logfile: googleapi: Error 403: <service account> does not have storage.objects.create access to the Google Cloud Storage object., forbidden```

This was solved by manually adding the role `storage.admin` to the service account for the logs bucket.

### PR DESCRIPTION
**Pull request by pasha-gh**
Fix #1050

### CODE REVIEW

File: examples/tfengine/generated/org_foundation/cicd/main.tf [hunk 1]
```diff
@@ -64,6 +64,7 @@ locals {
     "roles/logging.configWriter",
     "roles/resourcemanager.projectCreator",
     "roles/resourcemanager.organizationAdmin",
+    "roles/storage.objectAdmin",
```
xingao267, line 67: Can we add this role to the logging bucket only instead of doing it at the folder/org level?

pasha-gh: Done.

templates/tfengine/components/cicd/main.tf
```diff
@@ -305,6 +305,16 @@ resource "google_storage_bucket_iam_member" "cloudbuild_state_iam" {
   ]
 }
 
+# IAM permissions to allow Cloud Build SA to access logs bucket.
+resource "google_storage_bucket_iam_member" "cloudbuild_logs_bucket_iam" {
+  bucket = var.logs_bucket
```
xingao267, line 310: we should use reference here so logs_bucket is created first before this resource is attempted

thomasfung-hk: @xingao267 should this be done for the other IAM permissions in the same file as well then?

xingao267: Depending on whether or not the dependent resource is created in the cicd recipe or not. For example, the state bucket is not created here, but in the `devops` recipe, and the `cicd` recipe takes that as an input, so there is no need to do that for the state bucket. Similarly for the billing account which is not something created in this recipe. However, if the resource is created here, then we should make sure the dependency is sorted out clearly to avoid race conditions.

thomasfung-hk (approved):

templates/tfengine/components/cicd/main.tf
pasha-gh: Done.

xingao267 (approved):

Issue #974, opened 2021-07-22
Add devops, groups and cicd recipe to tfengine unit/integration tests


xingao267:
so it's tested by a local `terraform plan`. There likely is some historical reason why it's not included at this moment.

xingao267:
Could also look into adding it to the integration test.

xingao267:
Currently the integration test uses `team.hcl`. The cicd template in the `foundation.hcl` used by `team.hcl` is currently using github as the source repo system. That needs to be changed to cloud source repository.

thomasfung-hk:
Added `devops` to `tfengine` integration test and updated build configurations

thomasfung-hk:
Added `groups` to the scope of this ticket since `cicd` is dependent on it.

Currently, `groups` and `cicd` have been added in the fork of the repo.

thomasfung-hk:
Added `groups` and `cicd` to integration test; merged changes.

PR #1065, opened 2021-11-18
Add devops to tfengine integration test

thomasfung-hk:
First half of Issue #974.

These changes add devops module to the integration test. In order to make the build work, some additional permissions should be given to the service account as well, to ensure it can build and tear down the Terraform state properly.

- `roles/resourcemanager.projectDeleter`
- `roles/serviceusage.serviceUsageAdmin`
- `roles/resourcemanager.lienModifier`

In addition, the service account needs to be able to view, create and delete groups in the folder.

`CUSTOMER_ID` also needs to be an environment variable for the groups in `devops` to be built.


thomasfung-hk:
Just added `_CUSTOMER_ID` to integration test build configurations.

xingao267 (approved):

PR #1066, opened 2021-11-19
Remove TODO(mpetkov)

MartinPetkov:
I no longer remember what this was supposed to mean, and either way it shouldn't be assigned to me directly; if someone can clarify the missing functionality here, it should be `TODO(<issue #>)`.


internal/terraform/json.go
```diff
@@ -202,7 +202,6 @@ func resolveExpression(expr map[string]interface{}, plan *plan) interface{} {
 // At the moment, it only handles "var.XXX", for references to variables.
 // The docs say not to parse the strings, but I don't see anywhere in the plan where these are directly used as keys, so I'm parsing them as strings.
 // https://www.terraform.io/docs/configuration/expressions.html#references-to-named-values
-// TODO(mpetkov): Find out how to canonically resolve all types of references and expressions.
```
xingao267, line 205: Is this func used by tfimporter only?

MartinPetkov: Yes it is.

xingao267 (approved):

PR #1068, opened 2021-11-19
Support terraform_addons in audit recipe.

xingao267:


templates/tfengine/recipes/audit.hcl
```diff
@@ -108,6 +108,12 @@ schema = {
         type = "string"
       }
     }
+    terraform_addons = {
+      description = <<EOF
+        Additional Terraform configuration for the cicd deployment.
```
pasha-gh, line 113: for the `audit` deployment.

pasha-gh (approved):
minor documentation comment

templates/tfengine/recipes/audit.hcl
xingao267: Good catch! Fixed.

thomasfung-hk:
What is the reason that we are restricting the provider versions to that range?

xingao267:
> What is the reason that we are restricting the provider versions to that range?

No specific reason, just an example to demonstrate the custom provider constraints functionality. I added a comment.

thomasfung-hk:
Looks like it failed the integration test while trying to create `ManagedZone`

thomasfung-hk (approved):

xingao267:
> Looks like it failed the integration test while trying to create `ManagedZone`

That flake has been around for a while actually. Never got a chance to look into that more. If it happens more often, we can spend some bandwidth to fix it.

PR #1073, opened 2021-11-29
Remove unencrypted Git protocol from documentation

thomasfung-hk:
The current README files and commented recipe paths recommend using the `git://` protocol, which will be deprecated: https://github.blog/2021-09-01-improving-git-protocol-security-github/

We are removing all suggestions to use `git://` and to simply use the link starting with `github.com`, which gets automatically interpreted as a fetch request over Git using HTTPS.


thomasfung-hk:
Looks like there are more commits than expected from a previous merge; regardless, the changed files only show a change in documentation.

xingao267:
@thomasfung-hk I usually just create a branch out of this repo instead of from the fork, and then `git fetch` then `git rebase` then `git checkout origin/master -b test` so your local feature branch is even with up-to-date remote master branch.

xingao267 (approved):

xingao267:
When you `Squash and merge`, make sure to delete irrelevant commit messages.

PR #1074, opened 2021-11-30
Changed plan & apply in tf-apply-{env} job

### LINKED ISSUE #1022: Plan & apply in tf-apply-{env} job
Hi all,

Is there a reason we run a `terraform plan` before the `terraform apply -auto-approve` in our `tf-apply-{env}` cicd jobs? Was just looking over the docs for terraform apply and it seems that by default it runs the plan on its own anyway? 
docs: https://www.terraform.io/docs/cli/commands/apply.html#usage
config: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/components/cicd/configs/tf-apply.yaml#L38

Further, if we did want to keep it as is, maybe the apply could reference the plan output so that we didn't create it again? This would also remove the chance of the state changing between plan & apply.

For example:
```
terraform plan -out tfplan
terraform apply -auto-approve "tfplan"
```

thomasfung-hk:
Closes #1022.

This updates the triggers built using `tf-apply.yaml` to output the speculative plan during `terraform plan` into a file, which is then used in `terraform apply` to ensure that the plan in the output is consistent with the actual plan applied.

These changes can be also seen as suggested on the Hashicorp Learn site regarding automating Terraform: https://learn.hashicorp.com/tutorials/terraform/automate-terraform#automated-terraform-cli-workflow


pasha-gh (approved):

xingao267:
The change LGTM. Two questions:

1. Have you tested this?
2. Is the plan still printed in the stdout?

thomasfung-hk:
1. Yes, I tested it on a `folder_foundation` project.
2. Yes, the plan is printed and displayed on the build log when I ran the trigger.

xingao267 (approved):

Issue #1062, opened 2021-11-15
Support labels in project component and recipes


xingao267:
What are labels: https://cloud.google.com/resource-manager/docs/creating-managing-labels
An example of existing resources that already support labels: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/components/resources/gke_clusters/main.tf#L54

PR #1079, opened 2021-12-08
Support labels in project

kvoss:
see #1062


examples/policygen/generated/forseti_policies/policies/templates/gcp_iam_restrict_service_account_key_age_v1.yaml
```diff
@@ -67,16 +67,23 @@ spec:
...
+          check_key_not_expired(key) {
           	expiry_time := time.parse_rfc3339_ns(lib.get_default(key, "validBeforeTime", "1900-01-01T01:00:006Z"))
+          	expiry_time >= 0
           	now := time.now_ns()
-          	check_key_not_expired := now < expiry_time
+          	now < expiry_time
           }
           
-          check_key_age(key, max_age) = check_key_age {
+          # Workaround for dates in the far future - https://github.com/open-policy-agent/opa/issues/4098
```
thomasfung-hk, line 77: Were the changes in this file generated from your changes in the HCL files?

kvoss: No, my changes did not touch policies. `scripts/regen.sh` updated this file.

thomasfung-hk (approved):

templates/tfengine/recipes/project.hcl
```diff
@@ -136,6 +136,13 @@ schema = {
             }
           }
         }
+        labels = {
```
xingao267, line 139: Follow the existing way of specifying resource labels? e.g. https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/templates/tfengine/recipes/resources.hcl#L83

examples/tfengine/generated/team/project_secrets/main.tf
```diff
@@ -52,6 +52,9 @@ module "project" {
   # It is a no-op when Kubernetes Engine API is not enabled in the project.
   grant_services_security_admin_role = true
   activate_apis                      = ["secretmanager.googleapis.com"]
+  labels = {
```
xingao267, line 55: Can we also add a label in the project block (https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/examples/tfengine/modules/team.hcl#L36) and make sure that label value can also get propagated here?

kvoss: I tested explicitly adding labels in the block. They were correctly merged with the labels defined at root.
There was no instance where explicitly defined label was not accompanied by one from the root.

Do we want to reference labels defined in a parent scope, and not only current or root scope?

kvoss: There is an issue in the current implementation.

Adding labels map in the project block propagates them down to the appropriate project template.
However, that merged map also propagates to subsequent project templates in the recipe that should not have inherited the map.

kvoss: Analyzing `examples/tfengine/team.hcl` with labels added to the `project_secrets`
https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/d955b6b5d8a58a8589d6716e325c1eab2e3cf211/examples/tfengine/modules/team.hcl#L36-L41

In `recipes/project.hcl`, `project` template configures `flatten` for the `project` block.
https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/d955b6b5d8a58a8589d6716e325c1eab2e3cf211/templates/tfengine/recipes/project.hcl#L165-L170

Next, the elevated fields (including `labels`) are merged onto `templateInfo.Data`, which is merged back onto local `data` map that was earlier merged with `conf.Data`.
https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/d955b6b5d8a58a8589d6716e325c1eab2e3cf211/internal/tfengine/tfengine.go#L166-L180

After that, the `labels` map of the root object contains unwanted project labels.

kvoss: 58fb136 adds a label in the project block and shows the proper propagation of that label.
57a1bbc makes a copy of root data map, before merging local data maps.

examples/tfengine/generated/team/project_secrets/main.tf
```diff
@@ -52,6 +52,10 @@ module "project" {
...
+    env     = "prod"
+    no-gdpr = "true"
```
xingao267, line 57: Can we use `type = "no-phi"` to be consistent with other labels?

internal/template/template.go
```diff
@@ -128,6 +128,20 @@ func MergeData(dst map[string]interface{}, src map[string]interface{}) error {
 	return mergo.Merge(&dst, src, mergoOpts...)
 }
 
+// CopyData creates a copy of the argument map src.
+// Nested maps are copied recursively.
+func CopyData(src map[string]interface{}) map[string]interface{} {
```
xingao267, line 133: Please add some tests for this func in `template_test.go`.

internal/tfengine/tfengine.go
```diff
@@ -157,16 +157,12 @@ func dump(conf *Config, pwd, cacheDir, outputPath string, wantedTemplates map[st
...
 	outputPath = filepath.Join(outputPath, ti.OutputPath)
 
-	data := make(map[string]interface{})
+	data := template.CopyData(conf.Data)
 
 	// Make the schema available.
 	trimDescriptions(conf.Schema)
 	data["__schema__"] = conf.Schema
 
-	if err := template.MergeData(data, conf.Data); err != nil {
```
xingao267, line 166: Could you explain why MergeData does not work in this case for the `labels` map for the project recipe, but `labels` map for other resources, e.g. GKE works and other fields in the project recipe also works?

internal/template/template.go
```diff
@@ -128,6 +133,12 @@ func MergeData(dst map[string]interface{}, src map[string]interface{}) error {
 	return mergo.Merge(&dst, src, mergoOpts...)
 }
 
+// CopyData creates a deep copy of the argument map src.
+func CopyData(src map[string]interface{}) map[string]interface{} {
+	data := copier.Copy(src).(map[string]interface{})
```
xingao267, line 138: Since you are doing a type assertion here, we probably should also catch errors if the type of return value is not map[string]interface{} , e.g.

data, ok = ...
if not ok {
  # log error
}
return data

internal/template/template_test.go
```diff
@@ -91,6 +92,40 @@ func TestMergeData(t *testing.T) {
 	}
 }
 
+func TestCopyData(t *testing.T) {
+	nested := map[string]interface{}{
+		"c": []int{1, 2, 3},
+	}
+	addrNested := fmt.Sprintf("%p", nested)
```
xingao267, line 99: was it for your own debugging?

internal/template/template_test.go
```diff
@@ -91,6 +92,40 @@ func TestMergeData(t *testing.T) {
...
+				"a": nested,
+			},
+			want: map[string]interface{}{
+				"a": nested,
+			},
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			copy := CopyData(tc.input)
```
xingao267, line 117: got := CopyData(tc.input)

internal/template/template_test.go
```diff
@@ -91,6 +92,40 @@ func TestMergeData(t *testing.T) {
...
+			addrCopy := fmt.Sprintf("%p", copy["a"])
```
xingao267, line 118: We can delete this.

internal/template/template_test.go
xingao267: nvm saw address comparison below

kvoss: I wanted to extract the address of the example dictionary, and found this to be one way of getting that address

internal/template/template_test.go
xingao267, line 118: no need to create a var for this if the value is just used once

internal/template/template_test.go
xingao267, line 99: `addrOrginal` might be a more intuitive var name

internal/tfengine/tfengine.go
xingao267: Answered in https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/pull/1079#discussion_r774024687

examples/tfengine/generated/team/project_secrets/main.tf
```diff
@@ -52,6 +52,10 @@ module "project" {
   # It is a no-op when Kubernetes Engine API is not enabled in the project.
   grant_services_security_admin_role = true
   activate_apis                      = ["secretmanager.googleapis.com"]
+  labels = {
+    env  = "prod"
+    type = "no-phi"
```
thomasfung-hk, line 57: What is this label about?

kvoss: The `no-phi` indicates that there is no `Protected Health Information` in the `project_secrets`.

internal/template/template_test.go
kvoss: I removed using addresses from the test. Instead, I test for side-effect: compare copy with original before and after an external modification. If deep copy fails then objects do not differ after a modification.

internal/template/template_test.go
kvoss: I removed comparing addresses from the test.

internal/template/template.go
```diff
@@ -128,6 +133,18 @@ func MergeData(dst map[string]interface{}, src map[string]interface{}) error {
 	return mergo.Merge(&dst, src, mergoOpts...)
 }
 
+// errTypeAssertion captures a CopyData error if deep copy cannot be asserted to be a map[string]interface{}
+var errTypeAssertion = errors.New("CopyData could not assert data type")
+
+// CopyData creates a deep copy of the argument map src.
+func CopyData(src map[string]interface{}) (map[string]interface{}, error) {
+	if data, ok := copier.Copy(src).(map[string]interface{}); ok {
```
xingao267, line 141: The more common pattern is to check if it is not ok, and return the error in that case. And you don't need to create a separate way for that. So in this case:

```
if data, ok := copier.Copy(src).(map[string]interface{}); !ok {
  return nil, errors.New("copyData could not assert data type")
}
return data, nil
```

And as a nit, in inner error messages, use lower case as the start of the sentence (e.g. https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/master/internal/policygen/load.go#L57). It is because this error message can be appended to upper level error message and readability wise, having an upper case in it is not good. Usually only the out-most error message, for example, log.Errorf(...) in the top main function of the binary will have first latter capitalized as that is usually the start of the sentence.

See more at go/go-style/decisions#error-strings

xingao267: Also in non-test code, we normally don't directly mention the function name in the error message. Instead, interpret what the error is, or use "failed to ...". In this case, you can say something like "failed to assert output data type during data copy "

internal/template/template.go
```diff
@@ -128,6 +133,18 @@ func MergeData(dst map[string]interface{}, src map[string]interface{}) error {
...
+		return data, nil
+	}
+	// err := fmt.Errorf("CopyData failed to return the expected type; got %T", data)
```
xingao267, line 144: remove this line

internal/template/template_test.go
```diff
@@ -91,6 +91,33 @@ func TestMergeData(t *testing.T) {
...
+		nested := map[string]interface{}{
+			"c": []int{1, 2, 3},
+		}
+		outer := map[string]interface{}{
+			"a": nested,
+		}
+
+		got, err := CopyData(outer)
+		if err != nil {
+			t.Errorf(err.Error())
```
xingao267, line 105: `t.Errorf("CopyData: %v", err)`

internal/template/template_test.go
```diff
@@ -91,6 +91,33 @@ func TestMergeData(t *testing.T) {
...
+		}
+		if diff := cmp.Diff(outer, got); diff != "" {
+			t.Errorf("CopyData destination differs (-want +got):\n%v", diff)
```
xingao267, line 108: t.Errorf("copied data differs (-want +got):\n%v", diff)

internal/template/template_test.go
```diff
@@ -91,6 +91,33 @@ func TestMergeData(t *testing.T) {
...
+		}
+
+		// modify nested object
+		nested["b"] = 3
+
+		// deep copy (got) and outer map should differ now
```
xingao267, line 114: I think what you did before to compare the address is more straightforward than this.

internal/template/template.go
```diff
@@ -128,6 +133,15 @@ func MergeData(dst map[string]interface{}, src map[string]interface{}) error {
 	return mergo.Merge(&dst, src, mergoOpts...)
 }
 
+// CopyData creates a deep copy of the argument map src.
+func CopyData(src map[string]interface{}) (map[string]interface{}, error) {
+	data, ok := copier.Copy(src).(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Failed to assert output data type during data copy.")
```
xingao267, line 140: "failed to ..."

internal/template/template_test.go
```diff
@@ -91,6 +91,61 @@ func TestMergeData(t *testing.T) {
...
+				tc.outerConfig["innerConfig"] = tc.innerConfig
+			}
+
+			// Perfom copying test expecting no difference.
+			got, err := CopyData(tc.outerConfig)
+			if err != nil {
+				t.Errorf("CopyData: %v", err)
+			}
+			if diff := cmp.Diff(tc.outerConfig, got); diff != "" {
+				t.Errorf("Copied data differs (-want +got):\n%v", diff)
```
xingao267, line 134: "copied data"

internal/template/template_test.go
```diff
@@ -91,6 +91,61 @@ func TestMergeData(t *testing.T) {
...
+			}
+
+			// Modify inner config and expect difference between the created copy and the outer config.
+			if tc.innerConfig != nil {
+				tc.innerConfig["addedField"] = 3
+				if diff := cmp.Diff(tc.outerConfig, got); diff == "" {
+					t.Errorf("CopyData destination is expected to differ but it does not.\n")
```
xingao267, line 141: "copied data is expected to differ but it does not"

xingao267 (approved):
LGTM! Thanks!!

Issue #1083, opened 2021-12-13
Support reading from custom file in HCL configs


yeweidaniel:
Sometimes we need to provide a custom HL7 parser config to the the HL7 store, such as this example here: https://github.com/GoogleCloudPlatform/healthcare-data-protection-suite/blob/4276c1ae548ea7b872c1d996ad6ce72bf19ec91f/examples/tfengine/modules/team.hcl#L444 

This custom config is a JSON file that could be several thousand lines long. It would be impractical to copy paste the entire JSON file into a HCL config file, can we have a custom function that reads a separate config file, something like [templatefile](https://www.terraform.io/docs/language/functions/templatefile.html) in Terraform?

xingao267:
Fixed by https://github.com/GoogleCloudPlatform/solution-acceleration-toolkit/pull/1112

PR #1106, opened 2022-03-08
[prototype] support read hl7v2 parser config from file

xingao267:
#1083


xingao267:
Prototyping a solution. If continue with this, need to provide option to alternate between `schema` and `schema_file`.

xingao267:
> This prototype was deployed successfully via Terraform, but need to double check if the parser_config is applied correctly in CHC API in Cloud Console.

Update: verified that it it applied to the actual store successfully by running `gcloud healthcare hl7v2-stores describe hl7-store --dataset healthcare-dataset`.

PR #1126, opened 2022-04-29
Use google-beta for healthcare stores to use V3 HL7 parser

yeweidaniel:
V3 HL7 and V2 BQ export schema are beta only features, released in https://github.com/hashicorp/terraform-provider-google-beta/releases/tag/v4.17.0

### DIFF
--- a/templates/tfengine/components/resources/healthcare_datasets/main.tf
+++ b/templates/tfengine/components/resources/healthcare_datasets/main.tf
@@ -16,6 +16,7 @@ limitations under the License. */ -}}
 module "{{resourceName . "name"}}" {
   source  = "terraform-google-modules/healthcare/google"
   version = "~> 2.1.0"
+  provider = google-beta
 
   name     = "{{.name}}"
   project  = module.project.project_id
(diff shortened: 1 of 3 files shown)


templates/tfengine/components/resources/healthcare_datasets/main.tf
```diff
@@ -16,6 +16,7 @@ limitations under the License. */ -}}
 module "{{resourceName . "name"}}" {
   source  = "terraform-google-modules/healthcare/google"
   version = "~> 2.1.0"
+  provider = google-beta
```
xingao267, line 19: I don't recall offhand if `module` supports a `provider` field. My feeling is it will need to go in the cloud healthcare's CFT module and directly be added to the corresponding resource.

yeweidaniel: Updated CFT module to use google-beta. Thanks for the pointer!

yeweidaniel: Updated CFT module, thanks!

xingao267: Np. Once that is released, you will need to update the version of the module to use here.

yeweidaniel:
Reopened as https://github.com/GoogleCloudPlatform/solution-acceleration-toolkit/pull/1127

PR #1151, opened 2022-07-28
Remove Cloud Scheduler AppEngine Dependency app

GKozakjian:
Cloud Scheduler jobs for HTTP or Pub/Sub Targets can be deployed in multiple GCP Regions around the world and no longer require that an App Engine application be deployed.

Pin markdownlint-cli to version 0.31.1 to disable rule MD052, which is breaking presubmit_unit tests.

Release note: https://cloud.google.com/scheduler/docs/release-notes#February_11_2022


alextumanov (approved):
/gcbrun

PR #1164, opened 2022-10-25
Update cicd docker image to 1.6.2

fbeevikm:
Current version (1.7.1) has some bugs with the terraform version(1.3.0). So, changing the version to 1.6.2 which has terrraform version 1.2.9.


yeweidaniel (approved):
Any concerns for compatibility since we use 0.14.9? If we were using 1.2.9 before then feel free to ignore this.

PR #1175, opened 2023-01-04
Fix the presubmit unit test.

fbeevikm:
Update the docker image to developer-tools-light
Upgrade golangci-lint to v1.47.3

### DIFF
--- a/build/presubmit_int.yaml
+++ b/build/presubmit_int.yaml
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Terraform version used is 0.14.8.
+# Terraform version used is 1.2.9.
 substitutions:
-  _CFT_TOOLS_CONTAINER: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:8b0d85766044b0e4701729f9e70ee78dcecdb8688bd86915eb2a1cb00580ba1e"
+  _CFT_TOOLS_CONTAINER: "gcr.io/cloud-foundation-cicd/cft/developer-tools-light@sha256:e4ef6a7cb46ca110b97eb5fb7134fb9f3720b98560436792a4199c77e739dca0"
 
 timeout: 3600s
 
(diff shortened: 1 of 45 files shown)


build/presubmit_int.yaml
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Terraform version used is 0.14.8.
+# Terraform version used is 1.2.9.
 substitutions:
-  _CFT_TOOLS_CONTAINER: "gcr.io/cloud-foundation-cicd/cft/developer-tools@sha256:8b0d85766044b0e4701729f9e70ee78dcecdb8688bd86915eb2a1cb00580ba1e"
+  _CFT_TOOLS_CONTAINER: "gcr.io/cloud-foundation-cicd/cft/developer-tools-light@sha256:e4ef6a7cb46ca110b97eb5fb7134fb9f3720b98560436792a4199c77e739dca0"
```
yeweidaniel, line 17: Out of curiosity, how did you find which image to use and what's the difference in developer-tools-light?

templates/tfengine/components/terraform/main/main.tf
```diff
@@ -40,7 +40,7 @@ terraform {
     google-beta = ">= 3.0"
     {{end -}}
     {{if not $hasKubernetes -}}
-    kubernetes  = "~> 1.0"
+    kubernetes  = "~> 2.10"
```
yeweidaniel, line 43: Is this because the new Terraform version no longer supports 1.0 provider?

yeweidaniel (approved):

build/presubmit_int.yaml
fbeevikm: I had updated the cicd to use this image for tf-apply and it worked fine. This one has terraform and cloud sdk. Deveoper tools has other packages installed too. I don't think we are using that here.

templates/tfengine/components/terraform/main/main.tf
fbeevikm: Yes.

PR #1191, opened 2023-11-04
add support for lastUpdatedPartitionConfig

JosieLi-Google:


examples/tfengine/generated/team/project_data/main.tf
```diff
@@ -170,9 +170,8 @@ module "healthcare_dataset" {
           bigquery_destination = {
             dataset_uri = "bq://example-prod-data.${module.one_billion_ms_dataset.bigquery_dataset.dataset_id}"
             schema_config = {
-              recursive_structure_depth = 3
```
yeweidaniel, line 173: Is this removal intended?

yeweidaniel (approved):

examples/tfengine/generated/team/project_data/main.tf
JosieLi-Google: These fields are removed by running ./scripts/regen.sh. In my opinion, these fields should not be removed. I think the declaration of these fields are missing in some other files, which causes ./scripts/regen.sh to remove these codes. Let me investigate into this and I will re-run the unit test and integration test.

JosieLi-Google: These fields are added back by adding the declaration in templates/tfengine/components/resources/healthcare_datasets/main.tf and running ./scripts/regen.sh.

yeweidaniel (approved):

PR #1194, opened 2023-12-15
correct lastUpdatedPartitionConfig as optional

JosieLi-Google:


templates/tfengine/components/resources/healthcare_datasets/main.tf
```diff
@@ -103,9 +103,11 @@ module "{{resourceName . "name"}}" {
             schema_config = {
               recursive_structure_depth = "{{$v.bigquery_destination.schema_config.recursive_structure_depth}}"
               schema_type = "{{$v.bigquery_destination.schema_config.schema_type}}"
+              {{if has . "last_updated_partition_config" -}}
```
yeweidaniel, line 106: I dont think this work because . refers to the root level data object, but you want to look for within v. I think you can replace the whole block with 

`{{hclsField $v.bigquery_destination.schema_config "last_updated_partition_config" -}}

yeweidaniel (commented):
Can you test it by pointing your local scripts to your own fork's commit, then running tfengine with and without last_updated_partition_config?

yeweidaniel (approved):

